PR #1, opened 2020-06-23
Poll

lxcyqx:
Add ability to load poll options with checkbox as well as add new polling option

currently there is duplicate code in poll.js file for adding and loading polling options. I will get rid one of them eventually but currently both are needed to show functionality for loading and adding since it is not yet connected to database


lxcyqx:
> @googlebot I signed it!

lxcyqx:
> We found a Contributor License Agreement for you (the sender of this pull request), but were unable to find agreements for all the commit author(s) or Co-authors. If you authored these, maybe you used a different email address in the git commits than was used to sign the CLA ([login here](https://cla.developers.google.com/) to double check)? If these were authored by someone else, then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.
> In order to pass this check, please resolve this problem and then comment `@googlebot I fixed it.`. If the bot doesn't comment, it means it doesn't think anything has changed.
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fgoogleinterns%2Fstep100-2020%2Fpull%2F1) for more info**.

lxcyqx:
> @googlebot I fixed it.

lxcyqx:
> @googlebot I signed it!

lxcyqx:
> @googlebot I fixed it.

capstone/src/main/java/com/google/sps/Objects/Option.java
```diff
@@ -5,22 +5,22 @@
...
+    this.userVotes = new ArrayList<String>();
   }
 
   public String getText() {
     return this.text;
   }
 
-  public void addVote(int id) {
-    this.votes.add(id);
+  public void addVote(String id) {
```
g-asweet, line 19: `addVote(String id)` On its own, is not immediately clear.

I would recommend renaming the parameter `userId`

capstone/src/main/java/com/google/sps/Objects/Option.java
```diff
@@ -5,22 +5,22 @@
 public final class Option {
 
   private final String text;
-  private ArrayList<Integer> votes;
+  private ArrayList<String> userVotes;
```
g-asweet, line 8: How are you handling making sure one user can only vote once?

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
```diff
@@ -0,0 +1,34 @@
...
+    option2.addVote("1");
+    Option option3 = new Option("Option 3");
+    option3.addVote("3");
+    List<Option> options = new ArrayList<Option>();
+    options.add(option1);
+    options.add(option2);
+    options.add(option3);
+    String json = new Gson().toJson(options);
+    response.setContentType("application/json");
+    response.getWriter().println(json);
```
g-asweet, line 32: This is perfect MVP code. I will not nitpick until you're moving forward with a more robust solution :)

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -0,0 +1,137 @@
...
+  fetch("/poll")
+    .then(response => response.json())
+    .then(options => {
+      options.forEach(option => {
+        createOptionElementLoad(option);
+      });
+    });
+}
+
+function createOptionElementLoad(option) {
```
g-asweet, line 14: This is already 55 lines long, can you maybe break it into a couple of functions?

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -0,0 +1,137 @@
...
+  const votesLabel = document.createElement("div");
+  votesLabel.setAttribute("class", "votes-label");
+  const votesString = numVotes === 1 ? "task" : "tasks";
+  votesLabel.innerText = `${numVotes} ${votesString}`;
+  optionContainer.appendChild(votesLabel);
+  //Append each option to outer options container
+  optionsContainer.appendChild(optionContainer);
+}
+
+function createOptionElement(text) {
```
g-asweet, line 72: This is where a templating language can help out. IDK what you're plan is, but maybe consider the following (maybe not ALL within this CL):
- Try moving all styles into your .css.
- Add classnames into this logic to access your styles.
- Consider building this dom structure as a string where you can inject the values you care about. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals 
```javascript
let person = 'Mike';
let age = 28;

function myTag(strings, personExp, ageExp) {
  let str0 = strings[0]; // "That "
  let str1 = strings[1]; // " is a "

  // There is technically a string after
  // the final expression (in our example),
  // but it is empty (""), so disregard.
  // let str2 = strings[2];

  let ageStr;
  if (ageExp > 99){
    ageStr = 'centenarian';
  } else {
    ageStr = 'youngster';
  }

  // We can even return a string built using a template literal
  return `${str0}${personExp}${str1}${ageStr}`;
}

let output = myTag`That ${ person } is a ${ age }`;

console.log(output);
```
- Consider moving this into an HTML file and seeing if you can do the same thing.

capstone/target/portfolio-1/index.html
```diff
@@ -2,14 +2,14 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <title>My Portfolio</title>
-    <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
+    <title>AppName</title>
+    <link rel="stylesheet" href="styles/style.css">
```
g-asweet, line 6: Try to proactively scope your .css into targeted files, using a shared style.css file for only styles which are shared across more pages, maybe even only which are shared on ALL pages.

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/Objects/Option.java
lxcyqx: I want to make it so that people can vote on multiple options, kind of like how Facebook polling works

g-asweet: Even so, with an array of strings per option, what's preventing one user from voting multiple times for one option?

lxcyqx: I was thinking that since Option has a list of userVotes, then I can search for the id of the current user and if the id is contained in that list, then I can set the checkmark as checked

PR #2, opened 2020-06-23
Get basic login functionality set up

himaniyadav:
Added basic login functionality and a very basic UI:
- LoginServlet.java creates and returns a link that users can use to sign in with. Uses the Users API.
- Auth.js will contain the code to handle login requests.
- Currently, there is only a sign in link. No Register/Login toggle.

Here is a screenshot of what the page currently looks like: https://screenshot.googleplex.com/gSCpPCuGEDy

### DIFF
--- a/capstone/src/main/java/com/google/sps/Objects/User.java
+++ b/capstone/src/main/java/com/google/sps/Objects/User.java
@@ -6,18 +6,30 @@ public final class User {
 
   private final String userId;
   private final String name;
-  private final String firstname;
+  private final String firstName;
+  private final String lastName;
   private final String email;
+  private final int phoneNumber;
   private final ArrayList<Badge> badges;
   private final ArrayList<String> groups;
   private final ArrayList<String> interests;
 
-  public User(String userId, String firstname, String lastname, String email, 
+  public User(String userId, String firstName, String lastName, String email) {
+    this.userId = userId;
+    this.name = firstName + " " + lastName;
+    this.firstName = firstname;
+    this.lastName = lastName;
+    this.email = email;
+  }
+
+  public User(String userId, String firstName, String lastName, String email, int phoneNumber,
       ArrayList<Badge> badges, ArrayList<String> groups, ArrayList<String> interests) {
     this.userId = userId;
-    this.name = firstname + " " + lastname;
-    this.firstname = firstname;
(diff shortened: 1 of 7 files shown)


capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -4,24 +4,26 @@
...
+  private final String firstname;
   private final String email;
   private final ArrayList<Badge> badges;
-  private final ArrayList<Group> groups;
+  private final ArrayList<String> groups;
   private final ArrayList<String> interests;
 
-  public User(int userId, String name, String email, ArrayList<Badge> badges,
-      ArrayList<Group> groups, ArrayList<String> interests) {
+  public User(String userId, String firstname, String lastname, String email,
```
anikabagga: we could add get/set methods for first name, last name, phone number, interests

g-asweet: I would also recommend camel casing `firstName`, `lastName`

capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -47,6 +71,20 @@ public String getEmail() {
...
+    this.firstName = firstName;
+    this.name = firstName + " " + getLastName();
+  }
+
+  public void setLastName (String lastName) {
+    this.lastName = lastName;
+    this.name = getFirstName() + " " + lastName;
+  }
+
+  public void setPhoneNumber(int newPhoneNumber) {
```
g-asweet: I'm going to highly recommend against an int for phone number storage. You'll likely want a string that you format to your exact specification, as phone numbers are less dry and cut than you'd expect with area codes, international codes, etc. Additionally an int does not have enough bits to represent all phone numbers.

But it's generally agreed a string is better than a Long for this purpose
https://stackoverflow.com/questions/3483156/whats-the-right-way-to-represent-phone-numbers

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/Objects/User.java
himaniyadav: I see, that makes sense, thanks!

g-asweet (approved):

PR #3, opened 2020-06-23
Poll

lxcyqx:
Add ability to load poll options with checkbox as well as add new polling option

currently there is duplicate code in poll.js file for adding and loading polling options. I will get rid one of them eventually but currently both are needed to show functionality for loading and adding since it is not yet connected to database


lxcyqx:
> @googlebot I fixed it.

PR #5, opened 2020-06-24
polling functionality

lxcyqx:
Ability to load polling data and add new option
Currently there is duplicate code in the javascript file for loading and adding polling options but I will get rid of one of them. They are there to show that loading and adding works for now without being connected to database


anikabagga (approved):

g-asweet:
The entire `capstone/bin/.*` directory is binaries that get generated and should be in the `.gitignore`

g-asweet (changes requested):
Please add `capstone/bin/.*` to the `.gitignore` and remove all files that were added to the repo in the `/bin/` directory, ideally never uploading the `/bin/` directory itself.

Here's a bunch of stuff you can add to the gitignore to avoid uploading unnecessary files and binaries:
https://gist.github.com/sharno/6658bfbbc0ec6b3b2c8c

```
*.class

# Mobile Tools for Java (J2ME)
.mtj.tmp/

# Package Files #
*.jar
*.war
*.ear

# virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
hs_err_pid*

# netbeans
nbproject/private/
build/
nbbuild/

# the dist file could make some problems
# dist/
nbdist/
nbactions.xml
nb-configuration.xml
.nb-gradle/

# Eclipse
.classpath
.project
.settings/

# Intellij
.idea/
*.iml
*.iws

# Mac
.DS_Store

# Maven
log/
target/
pom.xml.tag
pom.xml.releaseBackup
pom.xml.versionsBackup
pom.xml.next
release.properties
dependency-reduced-pom.xml
```

I don't know that this will work as-is with you guys having a capstone/ folder, you may want to move all of that out or just add a .gitignore INTO the capstone folder instead.

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -0,0 +1,137 @@
...
+  let text = option["text"];
+
+  const optionsContainer = document.getElementById("options-container");
+  const optionContainer = document.createElement("option-container");
+  optionContainer.setAttribute("class", "option-container");
+  //Add challenge name
+  const challengeName = document.createElement("p");
+  const challengeText = document.createTextNode(text);
+  challengeName.appendChild(challengeText);
+  optionContainer.appendChild(challengeName);
```
himaniyadav, line 26: Even if you do end up getting rid of the duplicate code, I think it might make sense to break out some of these bits into their own functions for readability, so that whichever one of these functions you choose to keep isn't too long. (And if you decide that you need both of the createOptionElement functions, you can still avoid duplicate code by breaking up parts into their own sections.)

himaniyadav (commented):
Looks good, just make note of the target/ and bin/ files to remove

capstone/bin/.project
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
```
g-asweet, line 2: Make sure to remove all files in the /bin directory before submitting

g-asweet (changes requested):

PR #6, opened 2020-06-24
Create Post UI

anikabagga:


g-asweet:
It looks like you have to sign that agreement and then reply with `@googlebot I signed it!` before you can merge any code in

capstone/target/maven-archiver/pom.properties
```diff
@@ -1,5 +1,5 @@
 #Generated by Maven
-#Mon Jun 22 17:44:04 UTC 2020
+#Wed Jun 24 07:26:49 UTC 2020
```
g-asweet: @anikabagga @himaniyadav @lxcyqx Is this a file that needs to be updated? Could it be added to the `.gitignore`?

g-asweet: Could you all look together through the list of files that are being added and make sure gitignore any that aren't necessary?

capstone/target/maven-status/maven-compiler-plugin/compile/default-compile/inputFiles.lst
```diff
@@ -1 +1,10 @@
-/home/anikabagga/capstone-project/capstone/src/main/java/com/google/sps/servlets/DataServlet.java
+/home/anikabagga/step100-2020/capstone/src/main/java/com/google/sps/Objects/User.java
```
g-asweet: This file appears like it shouldn't be uploaded, since it directly references /home/anikabagga/...

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.sps.Objects.Comment;
+
+@WebServlet("/post")
+public class PostDataServlet extends HttpServlet {
+
+   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    // Receives submitted post 
+    long timestamp = System.currentTimeMillis();
+    User author = null;
```
himaniyadav: It might make sense for this to end up being the a String containing the user id, from which we can get the user's name, since we don't necessarily need all of the user's data to be loaded with the post. I think it's fine for now though.

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
```diff
@@ -0,0 +1,66 @@
...
+    String postText = "here is some text";
+    String challengeName = "Challenge Name";
+    String img = "";
+    ArrayList<User> likes = new ArrayList<>();
```
himaniyadav: Same comment as above, I think it would make sense if this ends up being a list of Strings containing user ids.

capstone/target/maven-archiver/pom.properties
himaniyadav: Yeah I've added target/ to the gitignore but I think some files had been added already prior to me doing that. So we should go back and delete everything under target/.

himaniyadav (commented):
Looks good, just make note of the target/ files to delete.

anikabagga:
@googlebot I signed it!

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
anikabagga: sounds good, changed it to a String

lxcyqx:
It looks like the CLA bot is still blocking your commits even though you signed the CLA. What I did was go on the link that the bot linked for "Googlers go here for more info" and checked what email was being used for each commit. It's likely that there is another email on record

capstone/target/maven-status/maven-compiler-plugin/compile/default-compile/inputFiles.lst
anikabagga: removed target folder

capstone/target/maven-archiver/pom.properties
anikabagga: removed target folder

g-asweet (approved):

g-asweet (approved):

capstone/src/main/webapp/group.html
```diff
@@ -3,13 +3,24 @@
...
     <title>Group</title>
   </head>
-  <body></body>
+
+  <body>
+    
+    <div class="group-content">
+
+      <div class="group-header">
+      </div>
```
nicorod-google: nit: Make this one line since it doesn't have any children
```suggestion
      <div class="group-header"></div>
```

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
```diff
@@ -0,0 +1,93 @@
...
+
+   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    // Receives submitted post 
+    long timestamp = System.currentTimeMillis();
+    String authorId = "Jane Doe";
+    String postText = request.getParameter("post-input");
+    String challengeName = "Challenge Name";
+    String img = "";
+    ArrayList<String> likes = new ArrayList<>();
```
lxcyqx, line 73: I think likes should be the id of the users who have liked the post right? In that case, I think datastore returns a long for id

capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -46,15 +46,15 @@ public String getImg() {
...
   }
 
   public void addComment(Comment newComment) {
     this.comments.add(newComment);
   }
 
-  public void addNewLike(User newLike) {
-    this.likes.add(newLike);
+  public void addNewLike(String newLikeUserId) {
+    this.likes.add(newLikeUserId);
```
nicorod-google: I see that we are baking in a lot of assumptions with the method and parameter names which makes someone using this method question:
- When can I use this method i.e. Do I need to make sure that I'm only using it for new likes? If so, how do I go about doing that, especially since the method itself seems to be implying that it will take care of that for us?
- How does a newLikeUserId differ from a userId? Are they the same? Is a newLikeUserId a type of userId?

Importantly, I want to address these questions. A Like being added to a Post shouldn't enforce that only new likes can be added, rather the method itself should do be enforcing that. Otherwise you'll need some other method that makes sure that only a new like are being passed in.

I'd like you to change this method so that it is able to add a Like, regardless of whether a User has already liked the Post or not, and guarantee that only a User who has not liked that Post before be included.

capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -5,28 +5,31 @@
...
 
   public User(String userId, String firstName, String lastName, String email) {
     this.userId = userId;
     this.name = firstName + " " + lastName;
-    this.firstName = firstname;
+    this.firstName = firstName;
     this.lastName = lastName;
     this.email = email;
+    this.badges = new ArrayList<>();
+    this.groups = new ArrayList<>();
```
nicorod-google: This will be a similar comment to the one I made in Post.java about addNewLike, except this is about addGroup. Github is not allowing me to comment on addGroup however since you didn't make changes to it in this PR. I understand you may not have been the person to make this file, however I think you can apply my comment above to this method as well.

Specifically, based on the comment "Adds a unique groupId string to the user's list of groups."
What is ensuring that a unique groupId is being added? This method should take care of that rather than it depend on the ID being passed in to be unique.

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
```diff
@@ -0,0 +1,93 @@
...
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import com.google.appengine.api.images.ImagesService;
+import com.google.appengine.api.images.ImagesServiceFactory;
+import com.google.appengine.api.images.ServingUrlOptions;
+import java.io.PrintWriter;
+import java.net.MalformedURLException;
+import java.net.URL;
+import com.google.sps.Objects.User;
+import com.google.sps.Objects.Post;
+import com.google.sps.Objects.Comment;
```
nicorod-google: I know with a lot of these you are preparing for the future, however until you are actually using the import you shouldn't include it. Make sure to only include the necessary imports

capstone/src/main/webapp/group.html
```diff
@@ -3,13 +3,30 @@
...
+
+      <div class="create-post">
+        <form id="post-form" method="POST" action="/post" >
+          <span class="user-img align-vertical"></span>
+          <textarea id="post-input" class="align-vertical" placeholder="Post a status" name="post-input" ></textarea>
+          <input class="post-btn align-vertical" type="submit"/>
+        </form>
+      </div>
+
+
```
nicorod-google, line 47: nit: remove extra space

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/User.java
himaniyadav: Hmm that's a good point. I'm a little confused because in my mind, the User class itself (and I think all of these classes) would only be used after retrieving things from the Datastore. Because we retrieve data from the database, use that data to create a User, Group, Post, etc object, and then convert that object to JSON which is the response to the client. So wouldn't ensuring the values are unique be the job of the server side code?

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const postAuthor = document.createElement('h3');
+  postAuthor.className = "post-author align-vertical";
+  postAuthor.innerText = post.authorId;
+  postDiv.append(postAuthor);
+
+  const postContent = document.createElement('p');
+  postContent.className = "post-content";
+  postContent.innerText = post.postText;
+  postDiv.append(postContent);
```
nicorod-google: You are doing a good job here with naming, but could you break each of these into their own helper functions that return the constructed element? That would help with readability and not needing to worry about knowing how a postAuthor etc. is created. I know you still need the constructed elements to add to postDiv, so you can do something like
`postDiv.append(createAuthor(post))`

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
```diff
@@ -0,0 +1,93 @@
...
+
+@WebServlet("/post")
+public class PostDataServlet extends HttpServlet {
```
g-asweet: MAYBE rename, maybe just put a comment right above this to disambiguate post/posts/data/etc

capstone/src/main/java/com/google/sps/servlets/BlobstoreServlet.java
```diff
@@ -0,0 +1,34 @@
+package com.googl.sps.servlets;
+
+import com.google.appengine.api.blobstore.BlobInfo;
+import com.google.appengine.api.blobstore.BlobInfoFactory;
+import com.google.appengine.api.blobstore.BlobKey;
```
nicorod-google, line 4: I don't think these are being used ATM so please remove them until you're using them

capstone/src/main/webapp/group.html
```diff
@@ -3,13 +3,35 @@
...
+  <body onload="loadPosts()">
+    
+    <div class="group-content">
+
+      <div class="create-post">
+        <form id="post-form" method="POST" action="/post" >
+          <span class="user-img align-vertical"></span>
+
+          <input type="text" id="post-input" class="align-vertical" name="post-input" value="Post a status">
+          <i class="fa-paper-plane" ></i>
```
nicorod-google: Nit: remove the extra space
```suggestion
          <i class="fa-paper-plane"></i>
```

capstone/src/main/webapp/group.html
```diff
@@ -3,13 +3,35 @@
...
+          <button type="submit"><i class="fa-paper-plane align-vertical" ></i><button>
```
nicorod-google: nit: remove the extra space
```suggestion
          <button type="submit"><i class="fa-paper-plane align-vertical"></i><button>
```

nicorod-google (changes requested):
I don't want to create any possible conflicts so I won't make you remove anything, but in the future instead of adding new features onto the open PR, could you create a new PR for those new additions? I know it can be a pain with multiple PRs, especially if they aren't chained together so then you need to copy paste the new changes, but it's a big help to have a smaller PRs so we can review them and get back to you faster

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
```diff
@@ -0,0 +1,93 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    Query query = new Query("Post").addSort("timestamp", SortDirection.DESCENDING);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+
+    List<Post> posts = new ArrayList<>();
+    for (Entity entity : results.asIterable()) {
+      
+      long timestamp = (long) entity.getProperty("timestamp");
```
g-asweet: Turning entities into their associated class should be a method on that class.

```java
Post post = Post.fromEntity(entity);
```

g-asweet: And maybe
```java
Entity entity = post.getEntity();
```

g-asweet: `post.getJson()`

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,48 @@
...
+  return postDiv;
+}
+
+// Gets URL for uploaded image
+function fetchBlobstoreUrlAndShowForm() {
+  fetch('/blobstore-handler')
+    .then((response) => {
+    return response.text();
+    })
+    .then((imageUploadUrl) => {
```
g-asweet: indent

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,48 @@
...
+  postContent.className = "post-content";
+  postContent.innerText = post.postText;
+  postDiv.append(postContent);
+
+  return postDiv;
+}
+
+// Gets URL for uploaded image
+function fetchBlobstoreUrlAndShowForm() {
+  fetch('/blobstore-handler')
```
g-asweet: maybe rename /images or /imagestore or something more descriptive

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/Objects/User.java
nicorod-google: Ideally the purpose of the Servlets should be just retrieving/storing data. These objects act as the middle layer between the database and frontend - they should be used to structure the data when we retrieve the data and modify the data until we are ready to store it. They should handle all of the complexities about the data. Otherwise, we are just using these objects to send in a RESPONSE and not utilizing the objects to their best potential. You could just as easily create a map of the data you want and send that as the RESPONSE to achieve the same goal.

I understand we haven't exactly been doing it like so for the portfolio and it might seem trivial, but the flow for adding a new group should be something like so:

[User] User clicks on Join Group button ->
[JavaScript] Join Group button performs a POST request ->
[Servlet] User entity is retrieved from database using user ID from the request ->
[Servlet] User object created from the User entity ->
[User Object] Add group to User object's set of Groups. If Group already exists, it is not added ->
[Servlet] Update User entity in database with the new set of Groups

This helps explain how to update an entity: https://cloud.google.com/appengine/docs/standard/java/datastore/creating-entities#Updating_entities

capstone/src/main/webapp/scripts/post.js
anikabagga: fixed

capstone/src/main/webapp/scripts/post.js
anikabagga: renamed to post-image-handler

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
anikabagga: added helper function to create and get entities

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
anikabagga: changed servlet name to GroupPostDataServlet

capstone/src/main/webapp/group.html
anikabagga: removed

capstone/src/main/webapp/group.html
anikabagga: removed

capstone/src/main/webapp/scripts/post.js
anikabagga: sounds good, I created helped functions for each component

anikabagga: Will also make sure to make separate and smaller PR for different features!

capstone/src/main/java/com/google/sps/servlets/PostDataServlet.java
anikabagga: sure I have removed them for now

capstone/src/main/java/com/google/sps/Objects/Post.java
anikabagga: Makes sense, I didn't consider the case in which a user can unlike and then like a post again. I have changed that method to make sure the same userId is not added multiple times. Also added a method to remove a userId from the likes arraylist

capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -5,28 +5,31 @@
...
+    this.interests = new ArrayList<>();
   }
```
nicorod-google: Are multiple User constructers intended here? If so, can you get rid of the duplicate code by calling the second constructer below here

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,58 @@
...
+  postContent.className = "post-content";
+  postContent.innerText = post.postText;
+  return postContent;
+}
+
+// Gets URL for uploaded image
+function fetchBlobstoreUrlAndShowForm() {
+  fetch('/post-image-handler')
+  .then((response) => {
+  	return response.text();
```
nicorod-google, line 64: nit: replace the tab with two spaces

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,58 @@
...
+  postDiv.className = "post-div";
+  postDiv.appendChild(createProfileImg(post));
+  postDiv.append(createAuthor(post));
+  postDiv.append(createPostText(post));
+  return postDiv;
+}
+
+// Create HTML element for post profile img
+function createProfileImg() {
+   const profileImgDiv = document.createElement('div');
```
nicorod-google: nit: remove the extra space

capstone/src/main/java/com/google/sps/servlets/BlobstoreKeyServlet.java
```diff
@@ -0,0 +1,22 @@
+import java.io.IOException;
+
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
```
nicorod-google: nit: remove this extra line

capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -46,15 +46,19 @@ public String getImg() {
...
   }
 
   public void addComment(Comment newComment) {
     this.comments.add(newComment);
   }
 
-  public void addNewLike(User newLike) {
-    this.likes.add(newLike);
+  public void addLike(String userId) {
+    if (!likes.contains(userId)) this.likes.add(userId);
```
nicorod-google: This is good, but could you instead switch this over to using a Set instead of an ArrayList? Similar to the conversation we had during the daily stand-up on Thursday

capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -4,18 +4,18 @@
 
 public final class Post {
 
-  private final User author;
+  private final String authorId;
```
nicorod-google, line 8: Question, why did we switch over to authorId instead of User? I thought during the standup on Thursday we had decided to user the Objects we were creating which will have the necessary ID and won't require you to get the data from the database because you'll already have it.

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/Post.java
nicorod-google: Ignore this comment for now. I'm looking into it based on Himani's comment in her PR about datastore only returning a list

capstone/src/main/java/com/google/sps/Objects/Post.java
anikabagga: Above Himani mentioned that we don't need all of the user's data to be loaded with each post, only the authorId and their name. I was planning on using the authorId to get their name when a post is submitted. Also open to changing it back to a User object.

capstone/src/main/java/com/google/sps/Objects/User.java
anikabagga: removed the other constructor

capstone/src/main/java/com/google/sps/Objects/Post.java
himaniyadav: Yeah sorry that's my bad, I made that comment before our discussion yesterday! Within the database we would just store the user's id but in the actual object it makes sense to just have it be a User object.

nicorod-google: Yeah if you could change it back that would be great!

capstone/src/main/java/com/google/sps/Objects/Post.java
nicorod-google: Just linking this incase you didn't see it: https://github.com/googleinterns/step100-2020/pull/8#discussion_r446358061
Instead of using an ArrayList change this to a HashSet, then you don't need to do the check if it contains, it will just not add it if it already does

capstone/src/main/webapp/group.html
```diff
@@ -3,17 +3,41 @@
...
     />
     <link rel="stylesheet" href="styles/group.css"></link>
-    <script src="scripts/poll.js"></script>
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <script src="scripts/post.js"></script>
+    <script src="scripts/poll.js"></script>
     <title>Group</title>
   </head>
 
+<<<<<<< HEAD
```
nicorod-google: nit: Make sure you remove this after the merge!

capstone/src/main/webapp/group.html
```diff
@@ -3,17 +3,41 @@
...
+        </form>
+      </div>
+
+      <div class="post-feed">
+        <div id="posts-container"></div>
+      </div>
+    </div>
+  </body>
+</html>
+=======
```
nicorod-google: nit: Make sure you remove this after the merge!

capstone/src/main/webapp/group.html
```diff
@@ -42,5 +66,4 @@
       </div>
     </template>
 
-  </body>
-</html>
```
nicorod-google: I think these were accidentally removed, add these back in

capstone/src/main/webapp/group.html
```diff
@@ -3,17 +3,41 @@
...
+           <button type="submit"><i class="fa fa-paper-plane align-vertical"></i><button>
+        </form>
+      </div>
+
+      <div class="post-feed">
+        <div id="posts-container"></div>
+      </div>
+    </div>
+  </body>
+</html>
```
nicorod-google: Remove these, they should be added down below at the end where they were accidentally removed

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/Post.java
anikabagga: That makes sense to me, changed likes to a hashset.

capstone/src/main/webapp/group.html
anikabagga: Oh my bad, I fixed the merge errors on my other branch but not on this one. I merged our HTML code, will fix formatting of UI later.

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    HashSet<String> likes = new HashSet<>();
+    ArrayList<Comment> comments = new ArrayList<>();
+
+    // Creates entity with submitted data and add to database
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(createPostEntity(timestamp, authorId, postText, challengeName, img, likes, comments));
+
+    // Redirect back to the HTML page.
+    response.sendRedirect("/group.html");
+  }
```
nicorod-google: I just wanted to ask, based on comments made by Lucy and Himani, does datastore support casting to a HashSet (line 57) and putting a HashSet (line 76/89)? This builds and doesn't produce any errors?

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
himaniyadav: I know Datastore supports putting a HashSet. I'm unsure about casting, I think that was the main issue when Lucy tried it.

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public Post getPostEntity(Entity entity) {
+    long timestamp = (long) entity.getProperty("timestamp");
+    String authorId = (String) entity.getProperty("authorId");
+    String postText = (String) entity.getProperty("postText");
+    String challengeName = (String) entity.getProperty("challengeName");
+    String img = (String) entity.getProperty("img");
+    List<String> likes = (ArrayList<String>) entity.getProperty("likes");
+    HashSet<String> likesSet = new HashSet<String>(likes);
```
nicorod-google: I apologize since I know I told you to do this, but this is the last change and then I'll approve. Since likes isn't being used anywhere, can you inline it when you create the HashSet and then change the name of the HashSet to likes? Ideally the variable names shouldn't need to have the datatype baked into the name.
```suggestion
    HashSet<String> likes = new HashSet<String>((ArrayList<String>) entity.getProperty("likes"));
```

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: No worries! I change it to an in-line statement. I was getting a null pointer exception so I added an additional check.

nicorod-google (approved):

g-asweet (approved):

g-asweet (approved):

g-asweet:
@googlebot I fixed it.

anikabagga:
@googlebot I fixed it.

PR #7, opened 2020-06-24
Poll UI

lxcyqx:
Poll UI
Ability to add new option and load previously added options from database
Changed javascript by factoring it out into template in HTML

### DIFF
--- a/capstone/src/main/webapp/group.html
+++ b/capstone/src/main/webapp/group.html
@@ -8,8 +8,25 @@
       rel="stylesheet"
       type="text/css"
     />
+    <link rel="stylesheet" href="styles/group.css"></link>
+    <script src="scripts/poll.js"></script>
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Group</title>
   </head>
-  <body></body>
+
+  <body onload="getPollOptions()">
+    <div class="poll-container">
+      <div class="inner-poll-container">
+          <div class="section-text">
+              Poll
+          </div>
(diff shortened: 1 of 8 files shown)


g-asweet (approved):

capstone/src/main/webapp/group.html
```diff
@@ -8,8 +8,42 @@
...
+    <div class="poll-container">
+      <div class="inner-poll-container">
+          <div class="section-text">
+              Poll
+          </div>
+          <div id="options-container"></div>
+          </div>
+        </br>
+        </br>
+        </br>
```
nicorod-google: What are you trying to achieve with these 3 line breaks? Could we perhaps achieve this using margin/padding?

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -0,0 +1,105 @@
+const BAR_WIDTH = "690";
+const BAR_HEIGHT = "55";
+
+function getPollOptions() {
+  fetch("/poll")
+    .then(response => response.json())
+    .then(options => {
+      options.forEach(option => {
+        // createOptionElementLoad(option);
```
nicorod-google: nit: remove this

nicorod-google (changes requested):

capstone/src/main/webapp/group.html
lxcyqx: Yes, I just fixed it to add margin instead, thanks!

nicorod-google (approved):

PR #8, opened 2020-06-25
Create Login UI and backend logic

himaniyadav:
- Build front end for logging in and signing up for the app. 
- Complete server-side logic for adding a user to Datastore upon signing up.
- Screenshots of new login screen UI:
image
image

**Note**: I'm not exactly sure if there is a point to having the "Register" section. The only difference is that the user inputs their first name and last name (which is something they can change in the app itself later on). The SignUpUser servlet is called regardless of whether a user pressed "Sign in" or "Sign up". Currently, the user input for first name and last name isn't properly functional anyway -- I think it's because the page refreshes after the user is redirected to the login screen, so the form values get cleared. If you guys have a workaround that would let the form values persist, that would be great. Though I didn't spend too much time trying to get it to work because I'm not sure if we should just get rid of the Register section altogether.


lxcyqx:
you can try getting rid of binary files using filter-branch

lxcyqx:
I think we can do something that Andrew mentioned, which would be not having a register tab. If it's the user's first time logging in, there should be some sort of alert letting them know that they are using their google email to log in for the first time.

capstone/src/main/webapp/scripts/auth.js
```diff
@@ -33,22 +33,60 @@ function authenticateUser() {
...
+    // Redirect to group page once logged in.
+    window.location.href = 'group.html';
   });
 }
+
+/* 
+ * Toggle login form.
+ */
+function login() {
+  let loginForm = document.getElementById("login");
```
lxcyqx: you can make it const since it seems like you're not changing the variable

himaniyadav:
Update: We decided that it wasn't necessary to have both a login and register method. Instead, we can display an alert or message if we detect that it is a user's first time signing into our web app. (Haven't implemented this yet.)

Here is what the new UI looks like: 
image

capstone/src/main/webapp/scripts/auth.js
```diff
@@ -33,22 +33,27 @@ function authenticateUser() {
  .then(response => response.json())
  .then((login) => {
     // need to set button's url to login url.
-    const loginUrl = login.url;
-    const loginButton = document.getElementById('login-btn');
-    loginButton.setAttribute('href', loginUrl);
+    const loginUrl = login.loginurl;
+    const loginButtons = document.getElementsByClassName("login-btn");
+    for(let i = 0; i < loginButtons.length; i++) {
```
g-asweet:
```javascript
for (
```

g-asweet: Only one login button.

capstone/src/main/webapp/scripts/auth.js
```diff
@@ -33,22 +33,27 @@ function authenticateUser() {
...
+      signUpUser();
+    }
  });
 }
 
 /*
- * Sign up the user. Placeholder code.
+ * Sign up the user for the app if they have not already signed up.
  */
 function signUpUser() {
```
g-asweet: rename this function to reflect that it is used for logged in and newly registered users.

capstone/src/main/webapp/scripts/auth.js
```diff
@@ -33,22 +33,27 @@ function authenticateUser() {
...
-  // Get the user's first name and last name value through the form.
-
   // Send a POST request to the signup servlet with the user's name as params.
-  fetch('/signup')
-  .then(response => response.json())
-  .then((login) => {
-    // Do something with the response.
+  fetch('/signup', {method: 'POST'})
```
g-asweet: this is also probably not named correctly

capstone/src/main/webapp/scripts/auth.js
```diff
@@ -33,22 +33,27 @@ function authenticateUser() {
...
+  .then(() => {
```
g-asweet, line 52: I'm wondering if a newly registered user should be sent to a place to update their name, or at least get a custom welcome message.

g-asweet (dismissed):

capstone/src/main/webapp/scripts/auth.js
nicorod-google: Himani did call that out as part of the update. I think for now we can just add a TODO and add that for a follow-up PR

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -30,16 +30,18 @@
...
-                      userService.getCurrentUser().getEmail(), logoutUrl);
+      json = String.format(
+        "{\"email\": \"%s\", \"loginurl\": \"%s\", \"logouturl\": \"%s\", \"loggedIn\": %b}",
+        userService.getCurrentUser().getEmail(), loginUrl, logoutUrl, true);
     } else {
-      json = String.format("{\"email\": \"%s\", \"url\": \"%s\"}", 
-                          "null", loginUrl);
+      json = String.format(
+        "{\"email\": \"%s\", \"loginurl\": \"%s\", \"logouturl\": \"%s\", \"loggedIn\": %b}",
+        "null", loginUrl, logoutUrl, false);
```
nicorod-google: I recommend talking to Lucy about this since I know she did it for her portfolio, but instead of creating a string representation of the object, which can get super messy and can get hard to make sure the syntax is correct, can we use an object or some other data structure that contains the necessary information, convert that to JSON, and you send that over?

These may help:
https://github.com/lxcyqx/step/blob/master/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java#L90
https://github.com/lxcyqx/step/blob/master/portfolio/src/main/java/com/google/sps/servlets/AuthenticationServlet.java#L63

capstone/src/main/java/com/google/sps/servlets/SignUpServlet.java
```diff
@@ -0,0 +1,72 @@
...
+
+@WebServlet("/signup")
+public class SignUpServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    UserService userService = UserServiceFactory.getUserService();
+    if (userService.isUserLoggedIn()) {
+      String email = userService.getCurrentUser().getEmail();
+      String userId = userService.getCurrentUser().getUserId();
```
nicorod-google: Since you aren't using these variables anywhere else other than to set the property, instead of creating the variables can you just inline it below?

capstone/src/main/java/com/google/sps/servlets/SignUpServlet.java
```diff
@@ -0,0 +1,72 @@
...
+      String first = userService.getCurrentUser().getNickname();
+      String last = "";
+
+      if (first.contains(" ")) {
+        String[] split = first.split(" ");
+        first = split[0];
+        last = split[1];
+      }
```
nicorod-google: I think I'm a little confused as to what is happening here. Does the UserService only store a user's name as one string? So you need to essentially find the first and last names?
Also, if you are getting the name from the UserService, do you users need to be able to change their name or could we just rely on the name from UserService? If the latter is the case, then we probably don't need to split first and last name

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
himaniyadav: Yes, that makes sense! Andrew and I mentioned this during our 1:1 too

capstone/src/main/java/com/google/sps/servlets/SignUpServlet.java
himaniyadav: Yes, the idea was to split up a user's first and last name. UserService doesn't necessarily return a user's actual name - if they don't have it set, it could just be their email address (which is why I did the check for if the name contains a space). So I would like for users to be able to manually change their name within the app. I think I will end up creating an extra page that users go to upon registering for the first time that will handle that. The reason I wanted first/last name separated was because on the feed page for our mock it greets the user with "Welcome back, [firstname]!" which seems friendlier and less formal than using their full name. 

But I'll try to clear up this code or maybe just leave the UserService name as is until the user goes in to manually change it.

capstone/src/main/webapp/scripts/auth.js
himaniyadav: Yes, I'll try to address this in my next PR!

capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -5,28 +5,31 @@
...
+  private String lastName;
   private final String email;
-  private final String phoneNumber;
+  private String phoneNumber;
   private final ArrayList<Badge> badges;
-  private final ArrayList<String> groups;
+  private final ArrayList<Group> groups;
   private final ArrayList<String> interests;
 
   public User(String userId, String firstName, String lastName, String email) {
```
nicorod-google: I made this same comment on Anika's PR, but are we intending to have two different User constructers? If so, you can reduce the duplicate code by calling the second constructer below here

capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -93,8 +96,12 @@ public void addInterest(String newInterest) {
...
-  /* Adds a unique groupId string to the user's list of groups. */
-  public void addGroup(String newGroupId) {
-    this.groups.add(newGroupId);
+  public void addGroup(Group newGroup) {
+    // Iterate through list, checking if Group already exists to ensure uniqueness.
+    // Could have used a HashSet here, but HashSets are not directly returned by Datastore,
+    // negating the search time complexity benefits.
+    if (!this.groups.contains(newGroup)) {
+      this.groups.add(newGroup);
+    }
```
nicorod-google: Since I've told Anika to makes these changes as well I would recommend letting her make the changes in order to avoid merge conflicts and revert this back to how it was, unless you need it for what you are working on.

Want to make sure everyone sees this: @himaniyadav @lxcyqx @anikabagga 
After looking into it, I think I have a solution, though please correct me if I'm wrong since y'all have more experience than me using Datastore: 
In Datastore, we said that a User will store the list of Group IDs, rather than a list of Groups correct? Regardless of using a Set/ArrayList/Map etc. we won't have the list of Group Entities when we make the request for a User. However, the list of Group IDs will help us get the Group Entities that we can then use to create Group objects. We can retrieve this list of IDs, then use the DatastoreService's get method (https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/DatastoreService.html#get_2) with that list of IDs to return a Map<Key, Entity> where key is the Group ID and Entity is the Group. Then using the Group Entity, create the Group object. You still need to store the Group objects in some data structure in the meantime while you modify the data so you can store them in a HashSet.

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -30,19 +32,20 @@
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     UserService userService = UserServiceFactory.getUserService();
 
-    String loginUrl = userService.createLoginURL("/group.html");
+    String loginUrl = userService.createLoginURL("/");
     String logoutUrl = userService.createLogoutURL("/");
```
nicorod-google: inline these where you are creating LoginResponse since you aren't using them elsewhere

capstone/src/main/webapp/scripts/auth.js
```diff
@@ -21,34 +21,38 @@
...
- * Fetch the authentication status of the user from the server.
  * Create the login url for the Users API.
+ * Fetch the authentication status of the user from the server.
  */
-function authenticateUser() {
+function createLoginUrl() {
   fetch('/login')
  .then(response => response.json())
  .then((login) => {
     // need to set button's url to login url.
```
nicorod-google: nit: "Need ..."

capstone/src/main/webapp/scripts/auth.js
```diff
@@ -21,34 +21,38 @@
...
-    const loginUrl = login.url;
-    const loginButton = document.getElementById('login-btn');
+    const loginUrl = login.loginUrl;
+    const loginButton = document.getElementById("login-btn");
     loginButton.setAttribute('href', loginUrl);
+    const loggedIn = login.loggedIn;
+    if (loggedIn) {
```
nicorod-google: No need to create the variable if you're just using it for this if statement

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/User.java
himaniyadav: That's a good point, at first I had overloaded it so that we could make a User object with limited information, but on second thought we don't actually need both of them.

capstone/src/main/java/com/google/sps/Objects/User.java
himaniyadav: Okay, I think that solution makes sense. So should I let Anika make the changes to User.java?

nicorod-google: Yes please, thank you!

nicorod-google: Yes please, thank you!

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -1,32 +1,25 @@
...
+  private String name;
+  private String firstName;
+  private String lastName;
   private final String email;
-  private final String phoneNumber;
-  private final ArrayList<Badge> badges;
-  private final ArrayList<String> groups;
+  private String phoneNumber;
+  private final LinkedHashSet<Badge> badges;
+  private final LinkedHashSet<Group> groups;
```
nicorod-google, line 15: Just to understand, is there a reason you chose LinkedHashSet instead of HashSet?

himaniyadav: because LinkedHashSet preserves the order and for something like badges I think it would be good to display them in the order/reverse order you received them

nicorod-google: Good reason! I'm unsure if datastore associates a timestamp with the Entities that are created/updated but I do remember y'all using a timestamp to sort the comments in your portfolio, so I recommend you do something similar here. If datastore has its own timestamp it's possible that it gets updated whenever that Entity is updated, if so I recommend you add a creation timestamp to the Badge object to remind y'all to add that when y'all create the Badges. 
Should Groups have the same ordering? If so, similarly, add a timestamp for that as well

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/User.java
himaniyadav: Hmm so should I change these to just a HashSet then?

himaniyadav: The issue I see with that is that even if we store a timestamp with the Badge or Group entities, and retrieve and add them to the User object in the desired order, by adding them to a HashSet we can no longer guarantee the order is preserved.

nicorod-google: Well I guess there are two things to consider here, Badges in the database and Badges outside of the database. If you can do similarly to what you did for the Comments for your portfolio where you were able to sort them by the timestamp, then a LinkedHashSet is likely not necessary because the data will already be sorted, though the LinkedHashSet would help maintain the order while you do things with the data outside of the database. Also I'm not sure if updating a Badge entity modifies the related timestamp (is the timestamp from datastore or is it an attribute of the Badge entity?). If the timestamp isn't modified then just do what you had done before with the Comments. If you feel like you'd need to maintain the order after you retrieve the sorted list from the database, then continue to use the LinkedHashSet, otherwise a simple HashSet works

nicorod-google: Sorry I didn't mean to confuse you between the Linked/HashSet. I think LinkedHashSet works and I agree with your comment, I was just asking about the timestamp because I don't know how the timestamp itself works.

himaniyadav: Oh, well I'm not sure how Anika and Lucy did it but the timestamp I used doesn't come from Datastore. I don't think Datastore stores a timestamp. I used System.currentTimeMillis() to get the time as a long.

nicorod-google: That should work for here as well! Just add a timestamp attribute to the Badge and Group objects, then I'll approve :-)

nicorod-google (approved):

himaniyadav:
Wait oh no I can't merge it until Andrew approves the changes :(
Edit: Nevermind, I addressed all the changes so hopefully dismissing his review is okay.

PR #9, opened 2020-06-25
Group poll

lxcyqx:
Poll functionality with css styling 

- Ability to check and uncheck each checkbox. 
- Number of votes increments and decrements accordingly. 
- Options are sorted in order of time and votes. The options with highest number of votes are at the top. Votes with equal amounts of votes are sorted in order of time. Newly added options (aka 0 votes initially) will be added to the bottom. 
- The length of inner darker green bar adjusts accordingly with the number of votes each option has. The option with the maximum number of votes has a full green bar and the other numbers are some calculated fraction of that
- Connect data with backend
- Checkboxes are checked accordingly depending on the user that is logged in

Screen Shot 2020-06-26 at 8 40 21 PM


capstone/src/main/java/com/google/sps/servlets/PollServlet.java
```diff
@@ -16,25 +16,32 @@
...
     for (Entity entity : results.asIterable()) {
       long id = entity.getKey().getId();
       String text = (String) entity.getProperty("text");
       List<String> votes = (ArrayList<String>) entity.getProperty("votes");
       Option option = new Option(id, text, votes);
       options.add(option);
+      // If current user voted for option, add to list of voted options
+      if (votes != null && votes.contains(userId)) {
+        votedOptions.add(id);
+      }
```
himaniyadav, line 60: I think I'm confused but why is voted options a list? Can't a user only vote for one option at a time?

capstone/src/main/java/com/google/sps/servlets/UpdateVotesServlet.java
```diff
@@ -0,0 +1,75 @@
...
+      votes = new ArrayList<String>();
+    }
+    /*
+     * If checkbox is unchecked and list of votes contains user, remove user id to
+     * list of votes for current option
+     */
+    if (checked.equals("false") && votes.contains(userId)) {
+      votes.remove(userId);
+    } else if (checked.equals("true") && !votes.contains(userId)) {
+      // If checkbox is checked and list of votes contains user, add user id to list
```
himaniyadav: Think you might mean to say "if list of votes does not contain user"

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
lxcyqx: This is to keep track of the options that the current user has voted for so that when the page is loaded, I use that information to mark certain checkboxes as checked

himaniyadav: I see, so to clarify, the current user is able to check multiple options?

lxcyqx: Yep

capstone/src/main/java/com/google/sps/Objects/response/PollResponse.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.List;
+
+import com.google.sps.Objects.Option;
+
+public final class PollResponse {
+
+  private final List<Option> options;
+  // List of ids of options for which current user voted
+  private final List<Long> votedOptions;
```
nicorod-google: I think I might be a little bit confused, what are you trying to achieve with the voteOptions here? From what I understand based on the comment it sounds like it is the Options for which a User has voted for. While a Poll is composed of Options, why does the Poll need to know what Options a User has voted for? I could be wrong so please clarify if so, but could this not be an attribute of a User instead of for the Poll?

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
```diff
@@ -16,25 +16,32 @@
...
     }
-
-    String json = new Gson().toJson(options);
+    PollResponse pollResponse = new PollResponse(options, votedOptions);
```
nicorod-google: Can you please break this up into a helper function?

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
nicorod-google, line 60: I think using a HashSet here could be really helpful. I believe my comment on Himani's PR should help with how to go about populating the HashSet and this would be a great example for others to follow: https://github.com/googleinterns/step100-2020/pull/8#discussion_r446358061

capstone/src/main/java/com/google/sps/servlets/UpdateVotesServlet.java
```diff
@@ -0,0 +1,75 @@
...
+    // Get id of changed checkbox
+    String optionIdString = request.getParameter("id");
+    // Get whether checkbox is currently checked
+    String checked = request.getParameter("checked");
+    long optionId;
+
+    try {
+      optionId = Long.parseLong(optionIdString);
+    } catch (NumberFormatException e) {
+      System.err.println("ERROR: Failed to parse to long");
```
nicorod-google: Rather than just printing an error message since this would really only be helpful for developers, could you send a response here as well?
This may be helpful for how to send an error: https://www.programcreek.com/java-api-examples/?class=javax.servlet.http.HttpServletResponse&method=sendError
Ideally after you do that, on the front-end you would then check the response to see if there is an error, and if there is then somehow display that to the user, however that is unnecessary ATM and just merely sending an error is fine

nicorod-google: You could also try this way to redirect the user to an error page: https://stackoverflow.com/questions/3885725/setting-both-error-message-and-body-from-servlet

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -10,12 +11,39 @@ function getPollOptions() {
     .then(options => {
       const optionsContainer = document.getElementById("options-container");
       optionsContainer.innerHTML = "";
-      options.forEach(option => {
+      options["options"].forEach(option => {
```
nicorod-google: If you are getting the "options" from options, does this variable perhaps represent something else other than an options?

capstone/src/main/java/com/google/sps/servlets/UpdateVotesServlet.java
```diff
@@ -0,0 +1,75 @@
...
+      System.err.println("ERROR: Could not get entity from datastore");
+      return;
+    }
+
+    /*
+     * Using ArrayList here because datastore will only return type ArrayList.
+     * Casting it to a HashSet will still have O(n) time complexity, so ArrayLists
+     * seem to be the best option in this scenario.
+     */
+    List<String> votes = (ArrayList<String>) optionEntity.getProperty("votes");
```
nicorod-google, line 116: I believe my comment above about the HashSet also applies here. While I agree that it would still have O(n) time complexity since you need to create the HashSet from the Map, that is an upfront cost that I believe is a good tradeoff since that would just occur once and then you can use the advantages of a Set, whereas with the List you'd be at its disadvantage while you use it

capstone/src/main/java/com/google/sps/servlets/UpdateVotesServlet.java
```diff
@@ -0,0 +1,75 @@
...
+
+@WebServlet("/update-votes")
+public class UpdateVotesServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get id of changed checkbox
+    String optionIdString = request.getParameter("id");
+    // Get whether checkbox is currently checked
+    String checked = request.getParameter("checked");
```
nicorod-google: Can you rename this to something along the lines of isOptionChecked? It's pretty ambiguous otherwise as to what checked means. 

Also, can you use Boolean's parseBoolean or valueOf methods to convert this to a boolean/Boolean? Checking if a string === "false"/"true" is not pretty.
https://www.javatpoint.com/java-string-to-boolean

capstone/src/main/java/com/google/sps/servlets/UpdateVotesServlet.java
```diff
@@ -0,0 +1,75 @@
...
+    /*
+     * If checkbox is unchecked and list of votes contains user, remove user id to
+     * list of votes for current option
+     */
+    if (checked.equals("false") && votes.contains(userId)) {
+      votes.remove(userId);
+    } else if (checked.equals("true") && !votes.contains(userId)) {
+      // If checkbox is checked and list of votes contains user, add user id to list
+      votes.add(userId);
+    }
```
nicorod-google: Can you break this up into helper methods please?

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -10,12 +11,39 @@ function getPollOptions() {
...
+ * @param {array} votedOptions
+ */
+function handleCheck(votedOptions) {
+  const checkboxes = document.querySelectorAll("input[type=checkbox]");
+  for (let i = 0; i < checkboxes.length; i++) {
+    let checkbox = checkboxes[i];
+    if (
+      votedOptions.find(function(checkboxId) {
+        return checkbox.id == checkboxId;
+      })
```
nicorod-google: Can you break this into a helper function and call the function here instead?

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -10,12 +11,39 @@ function getPollOptions() {
...
+    ) {
+      checkbox.checked = true;
+    } else {
+      checkbox.checked = false;
+    }
+  }
+  return;
+}
```
nicorod-google: Could you walk me through here as to what is happening?

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/response/PollResponse.java
lxcyqx: I discussed this with Andrew, but what I'm trying to do is create a PollResponse which is a wrapper class that contains information that I need that will be converted to json. I need to know which options the current logged in user has voted for so that when I load the page, I know which options to mark as checked. I want to get all the information I need with one fetch request from PollServlet, so even if I attributed this with User, I would still need another wrapper class to put together the poll and user information. In this case, I don't think I need to put this info with the User object because the User object also contains a lot of information that I will not be needing.

capstone/src/main/webapp/scripts/poll.js
lxcyqx: In the JSON string I have a field that's called options, which contains options data, so I have to index into that. I have now changed it to be called pollData for more clarity. It contains information for each option that should be loaded onto the screen as well as votedOptions (discussed in earlier comment) which indicates the options for which the current user has voted

capstone/src/main/webapp/scripts/poll.js
lxcyqx: I am going through the checkboxes to see which ones should be checked. votedOptions comes from the JSON which contains the ids of the checkboxes that the user has checked, so when I go through the list of checkboxes, I check to see which checkbox ids are in the votedOptions list so that I know which checkbox to mark as checked

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
lxcyqx: If I am only checking once whether votes.contains(userId) that is total of O(n) for the entire method. Converting it to a HashSet would also be O(n), so why would making it a HashSet be better? I am not going to be using this HashSet later on. It is only created and used within this method once

capstone/src/main/java/com/google/sps/servlets/UpdateVotesServlet.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /**
+   * Handles error for Java Servlet and displays that something went wrong.
+   *
+   * @param response    HttpServletResponse
+   * @param errorString error message
+   * @throws IOException exception thrown when cannot write to file
+   */
+  private void sendError(HttpServletResponse response, String errorString) throws IOException {
```
nicorod-google: You aren't actually using errorString

capstone/src/main/java/com/google/sps/servlets/UpdateVotesServlet.java
```diff
@@ -0,0 +1,152 @@
...
+     * Casting it to a HashSet will still have O(n) time complexity, so ArrayLists
+     * seem to be the best option in this scenario.
+     */
+    List<String> votes = (ArrayList<String>) optionEntity.getProperty("votes");
+    Set<String> votesSet;
+    /*
+     * If current option does not have any votes, initialize ArrayList.
+     */
+    if (votes == null) {
+      votes = new ArrayList<String>();
```
nicorod-google: Do you need votes here? It doesn't look like you use it anywhere else except in the else statement

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -1,37 +1,116 @@
...
+function getMaxVotes(pollData) {
+  let maxVotes;
+  // Gets the list of votes for the first poll option
+  let votesArray = pollData["options"][0]["votes"];
+  if (votesArray == null) {
+    maxVotes = 0;
+  } else {
+    maxVotes = votesArray.length;
+  }
+  return maxVotes;
```
nicorod-google: I think you can make this more succinct with a ternary operator:
 ```suggestion
  // Gets the list of votes for the first poll option
  let votesArray = pollData["options"][0]["votes"];
  return !!votesArray ? votesArray.length : 0;
```
In case you don't know the "!!" returns the boolean truthy/falsy association of a value: https://developer.mozilla.org/en-US/docs/Glossary/Truthy
Here it'll equate to true unless the array is null

File: capstone/src/main/webapp/scripts/poll.js (hunk 6 above)
Comment (nicorod-google) [reply capstone/src/main/webapp/scripts/poll.js]: Thank you, pollData makes more sense and makes it more clear 👍

File: capstone/src/main/java/com/google/sps/Objects/response/PollResponse.java (hunk 3 above)
Comment (nicorod-google) [reply capstone/src/main/java/com/google/sps/Objects/response/PollResponse.java]: Oh okay I see now, I didn't understand that this was the response, thank you for the clarification!

File: capstone/src/main/webapp/scripts/poll.js (hunk 11 above)
Comment (nicorod-google) [reply capstone/src/main/webapp/scripts/poll.js]: Ah okay wonderful, so just to make sure I have it correct: for each checkbox you check if that checkbox exists by look through the entire list of voted options? Is there any way you can think of to make this more efficient?

File: capstone/src/main/java/com/google/sps/servlets/UpdateVotesServlet.java [hunk 15]
```diff
@@ -0,0 +1,152 @@
...
+   * @param response    HttpServletResponse
+   * @param errorString error message
+   * @throws IOException exception thrown when cannot write to file
+   */
+  private void sendError(HttpServletResponse response, String errorString) throws IOException {
+    response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Cannot parse to long.");
+    response.getWriter().print("<html><head><title>Oops an error happened!</title></head>");
+    response.getWriter().print("<body>Something bad happened uh-oh!</body>");
+    response.getWriter().println("</html>");
+  }
```
nicorod-google: I like what you did here, but rather than making this specific to UpdateVotesServlet since that would require you to copy this method for all of the servlets you'd like to use this in, could you create an error handling class in a new file that has this method and you can call in whichever servlet you want? This would help directly right now since line 36 in PollServlet uses the previous way you were handling errors

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
nicorod-google: Thank you for being patient and I apologize for taking a while to respond, I've been thinking about this problem the entire day. 
Firstly regarding the runtime of this method, while you are right that converting to a HashSet would be O(n), since you need to iterate through each of the Option Entities, the runtime for the method as a whole would be more than O(n). However whether we convert votes to a HashSet or use the contains list method, both are O(n) complexity, therefore there isn't an apparent initial advantage to using one or the other. 
So, assuming your point about "it is only created and used within this method once" is correct and based on your comment in PollResponse above, then I think it is fine to leave it as is.

nicorod-google (changes requested):

lxcyqx:
@nicorod-google I think you have to sign the CLA

capstone/src/main/webapp/scripts/poll.js
lxcyqx: Oh yes! I couldn't get has() to work with sets but I was able to use the javascript version of hashmaps

nicorod-google:
@googlebot I fixed it.

capstone/src/main/java/error/ErrorHandler.java
```diff
@@ -0,0 +1,23 @@
...
+   * @throws IOException exception thrown when cannot write to file
+   */
+  public void sendError(HttpServletResponse response, String errorString) throws IOException {
+    response.sendError(HttpServletResponse.SC_BAD_REQUEST, errorString);
+    response.getWriter().print("<html><head><title>Oops an error happened!</title></head>");
+    response.getWriter().print("<body>Something bad happened uh-oh!</body>");
+    response.getWriter().println("</html>");
+  }
+
+}
```
nicorod-google, line 23: This is wonderful, thank you!

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -1,37 +1,113 @@
...
+ * @param {array} votedOptions
+ */
+function handleCheck(votedOptions) {
+  const checkboxes = document.querySelectorAll("input[type=checkbox]");
+  let votedOptionsSet = convertToSet(votedOptions);
+  for (let i = 0; i < checkboxes.length; i++) {
+    let checkbox = checkboxes[i];
+    markCheckbox(votedOptionsSet, checkbox);
+  }
+  return;
```
nicorod-google, line 64: Very nice job, I like this a lot better!

nicorod-google (approved):

PR #10, opened 2020-06-29
Show popup to newly registered users

himaniyadav:
Added the following functionality to log in screen:
- Detect if a user is new or has logged in before
- If a user is new, display a modal popup that allows users to fill in information
- Create User Entity based on entered information
- When finished, redirect to group page

Screenshot of new modal:
image

Note: I know we're currently discussing the need for timestamps on Group and Badge (which I added in my last PR). If we decide we don't need them, I can remove those variables and include the changes in this PR.

Also, since we just decided to add the feature, I built the UI without any Figma reference but I can make any changes needed

### DIFF
--- a/capstone/pom.xml
+++ b/capstone/pom.xml
@@ -22,7 +22,7 @@
       <artifactId>javax.servlet-api</artifactId>
       <version>4.0.1</version>
       <scope>provided</scope>
-    </dependency>    
+    </dependency>
     <dependency>
       <groupId>com.google.appengine</groupId>
       <artifactId>appengine-api-1.0-sdk</artifactId>
--- a/capstone/pom.xml
+++ b/capstone/pom.xml
@@ -22,14 +22,7 @@
       <artifactId>javax.servlet-api</artifactId>
       <version>4.0.1</version>
       <scope>provided</scope>
-    </dependency>
-    
-    <dependency>
-	  <groupId>com.google.code.gson</groupId>
-	  <artifactId>gson</artifactId>
-	  <version>2.8.6</version>
-	</dependency>
-    
+    </dependency>    
     <dependency>
       <groupId>com.google.appengine</groupId>
       <artifactId>appengine-api-1.0-sdk</artifactId>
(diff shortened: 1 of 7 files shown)


capstone/src/main/webapp/scripts/auth.js
```diff
@@ -39,19 +39,57 @@ function createLoginUrl() {
...
+function openRegisterModal() {
+  let modal = document.getElementById('register-modal');
+  modal.classList.toggle('show-modal');
+}
+
+/*
+ * Sign up new user for the app.
+ */
+function createNewUser() {
+  const firstName = document.getElementById('first').value;
```
lxcyqx: for the required fields such as "First name *" are you checking that if the user doesn't type anything then they can't submit?

himaniyadav: Yes, in the HTML they're marked as required so if they're empty it prompts the user to fill them out

himaniyadav: I looked again and it looks like there actually is an intermediate state where it's creating a UserEntity, even if the form is invalid. Though after it prompts the user to fill out the fields and user submits again, it ends up being overwritten. We probably still don't want that, so I'm including a check just in case.

capstone/pom.xml
```diff
@@ -22,14 +22,7 @@
...
       <scope>provided</scope>
-    </dependency>
-    
-    <dependency>
-	  <groupId>com.google.code.gson</groupId>
-	  <artifactId>gson</artifactId>
-	  <version>2.8.6</version>
-	</dependency>
-    
+    </dependency>
```
nicorod-google: nit: I think there is extra space at the end of the line

capstone/src/main/java/com/google/sps/servlets/CreateNewUserServlet.java
```diff
@@ -0,0 +1,83 @@
...
+    String first = request.getParameter("first");
+    String last = request.getParameter("last");
+    String phone = request.getParameter("phone");
+    ArrayList<String> interests = getInterests(request);
+    
+    UserService userService = UserServiceFactory.getUserService();
+    if (userService.isUserLoggedIn()) {
+      Entity userEntity = createUserEntity(userService, first, last, phone, interests);
+      DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+      datastore.put(userEntity);
```
nicorod-google: nit: inline the createUserEntity in the datastore.put() since you aren't using it anywhere else.

nicorod-google (changes requested):

nicorod-google (approved):

PR #11, opened 2020-06-29
Post Comments

anikabagga:
Completed comments feature:
- created comment container and submit comment button
- each post has a list of embedded comment entities 

Screenshot 2020-06-29 at 1 18 42 PM


capstone/src/main/java/com/google/sps/Objects/Comment.java
```diff
@@ -4,12 +4,12 @@
 
   private final long timestamp;
   private final String commentText;
-  private final User user;
+  private final String userId;
 
-  public Comment(long timestamp, String commentText, User user) {
+  public Comment(long timestamp, String commentText, String userId) {
```
himaniyadav, line 12: Wait, so why are we doing userId here instead of user? I understand in the Datastore itself that's how the data will be stored, but for a comment won't we probably need to know the name or at least profile pic of the author as well?

capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -1,21 +1,24 @@
 package com.google.sps.Objects;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 
 public final class Post {
 
-  private final User author;
+  private final long postId;
+  private final String authorId;
```
himaniyadav: Same question here, about a Post's author vs. authorId. I guess Post doesn't need to know all of the user data but at least their name, profile picture (eventually), and id, correct?

capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -1,21 +1,24 @@
...
   private final String postText;
   private final ArrayList<Comment> comments;
   private final String challengeName;
   private final long timestamp;
   private final String img;
-  private final ArrayList<User> likes;
+  private final HashSet<String> likes;
 
-  public Post(User author, String postText, ArrayList<Comment> comments, String challengeName, long timestamp, String img, ArrayList<User> likes) {
+  public Post(long postId, String authorId, String postText, ArrayList<Comment> comments, String challengeName, long timestamp, String img, HashSet<String> likes) {
```
g-asweet: this is now more than 80 chars wide

https://google.github.io/styleguide/javaguide.html#s4.5-line-wrapping

```java
  public Post(
    long postId,
    String authorId,
    ....
  ) {
    ...
  }
```

capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -46,15 +49,19 @@ public String getImg() {
...
 
   public void addComment(Comment newComment) {
     this.comments.add(newComment);
   }
 
-  public void addNewLike(User newLike) {
-    this.likes.add(newLike);
+  public void addLike(String userId) {
+    if (!likes.contains(userId)) this.likes.add(userId);
+  }
```
himaniyadav: I think since you changed likes to a HashSet, you won't need to do the extra "contains" check. A HashSet would already guarantee that values are unique

capstone/src/main/java/com/google/sps/servlets/BlobstoreServlet.java
```diff
@@ -0,0 +1,33 @@
...
+import java.net.URL;
+import java.util.List;
+import java.util.Map;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/post-image-handler")
+public class BlobstoreServlet extends HttpServlet {
```
himaniyadav: Maybe rename this file from BlobstoreServlet to PostImageServlet? Because I think eventually we'll need Blobstore on other pages/for other uses as well. Alternatively, if we do need to use Blobstore's uploadUrl for other functions, we could pass in the desired url destination as a request parameter, so we don't have to repeat code. In that case, I would rename the WebServlet from /post-image-handler to /image-upload-handler or something similar.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -0,0 +1,64 @@
...
+    String commentText = request.getParameter("comment-text");
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity postEntity = this.getPostFromId(response, postId, datastore);
+    ArrayList<EmbeddedEntity> allComments = (ArrayList<EmbeddedEntity>) postEntity.getProperty("comments");
+
+    // Create comment entity and add to comment arraylist for post
+    if(allComments == null) {
+      ArrayList<EmbeddedEntity> comments = new ArrayList<>();
+      comments.add(createCommentEntity(commentText, "user"));
```
himaniyadav, line 39: just checking to see if I understand, eventually the "user" parameter will be set to the user's id from UserService?

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,156 @@
...
+    Query query = new Query("Post").addSort("timestamp", SortDirection.DESCENDING);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+
+    List<Post> posts = new ArrayList<>();
+    List<Long> likedPosts = new ArrayList<>();
+    for (Entity entity : results.asIterable()) {
+      posts.add(getPostEntity(entity));
+      ArrayList<String> likes = (ArrayList<String>) entity.getProperty("likes");
+      if(likes != null && likes.contains(userId)) likedPosts.add(entity.getKey().getId());
```
himaniyadav: I think for style/readability purposes you should separate this out into multiple lines with braces. https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

himaniyadav: To clarify, is the only reason for getting the Post's likes as an ArrayList to check if the current user has liked it?

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,156 @@
...
+    return userPost;
+  }
+
+  private void createCommentObjectList(ArrayList<Comment> comments, Entity entity) {
+    ArrayList<EmbeddedEntity> commentEntitys = (ArrayList<EmbeddedEntity>) entity.getProperty("comments");
+    for(EmbeddedEntity comment: commentEntitys) {
+      comments.add(
+        new Comment((long) comment.getProperty("timestamp"), 
+        (String)comment.getProperty("commentText"), 
+        (String) comment.getProperty("userId"))
```
himaniyadav: nit: indent these last two lines so it's clearer they're a continuation of the first

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,156 @@
...
+      );
+    }
+  }
+
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Receives submitted post 
+    long timestamp = System.currentTimeMillis();
+    String authorId = "Jane Doe";
```
himaniyadav: I know this is just for testing, but are you planning on using authorId to mean the author's name or actual userId? If the former, I think you should change the variable name to reflect that

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,156 @@
...
+    Entity taskEntity = new Entity("Post");
+    taskEntity.setProperty("authorId", authorId);
+    taskEntity.setProperty("timestamp", timestamp);
+    taskEntity.setProperty("postText", postText);
+    taskEntity.setProperty("challengeName", challengeName);
+    taskEntity.setProperty("img", img);
+    taskEntity.setProperty("likes", likes);
+    taskEntity.setProperty("comments", createCommentEntities(comments));
+    return taskEntity;
+	}
```
himaniyadav, line 117: nit: this curly brace isn't in the right place 😅

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,156 @@
...
+      commentEntity.setProperty("timestamp", comment.getTimestamp());
+      commentEntity.setProperty("commentText", comment.getCommentText());
+      commentEntity.setProperty("userId", comment.getUser());
+      allComments.add(commentEntity);
+    } 
+    return allComments;
+  }
+
+   /** Returns a key that points to the uploaded file, or null if the user didn't upload a file. */
+  private String getUploadedFileUrl(HttpServletRequest request, String formInputElementName) {
```
himaniyadav, line 133: I know this is probably not finished yet, but where is this method used/going to be used? Also, how is the String parameter "formInputElementName" going to be used?

capstone/src/main/java/com/google/sps/servlets/UpdateLikesServlet.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+    return "";
+  }
+
+  private Entity getPostFromId(HttpServletResponse response, long postId, DatastoreService datastore) throws IOException {
+    try {
+      return datastore.get(KeyFactory.createKey("Post", postId));
+    } catch (EntityNotFoundException e) {
+      return null;
+    }
```
himaniyadav: I know technically the Entity should never be not found, so this probably won't ever happen, but maybe instead of returning null we should have some sort of error displayed to the user just in case something messes up. Otherwise the server will throw a null pointer exception but the user won't know what's gone wrong.

capstone/src/main/webapp/group.html
```diff
@@ -3,44 +3,52 @@
...
+  <body onload="init()">
+    
+    <div class="group-content">
+
+      <div class="group-header">
+        <div class="group-name-banner">
+          <span id="group-name" class="align-vertical">The 3 Musketeers + Others</span>
+        </div>
       </div>
-    </div>
```
himaniyadav: This is probably because y'all are working on the same file so merging and stuff with Git gets tricky but when you merge this into master Lucy's code won't be deleted, right?

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,193 @@
+function init() {
+  loadPosts();
+  uploadImage();
+  //fetchBlobstoreUrlAndShowForm();
+}
+
+function uploadImage() {
+  console.log("in upload image");
```
himaniyadav: nit: Get rid of these testing log statements if you don't need them anymore

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,193 @@
...
+  fetch('/group-post').then(response => response.json()).then((postsResp) => {
+    const posts = postsResp.posts;
+    const allPostsList = document.getElementById('posts-container');
+    allPostsList.innerHTML = '';
+    for (var i = 0; i < posts.length; i++) {
+      allPostsList.appendChild(createSinglePost(posts[i]));
+    }
+  }).then(() => {
+    var elements = document.getElementsByClassName('post-btn align-vertical comment-btn');
+    for (var i = 0; i < elements.length; i++) {
```
himaniyadav: nit: Use let and not var

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,193 @@
...
+      likeBtns[i].addEventListener("click", function() {
+        //this.classList.toggle("active");
+        likeToggled(this.id, true);
+      });
+    }
+  });
+}
+
+function likeToggled(likeId, liked) {
+  const postId = likeId.substring(0, likeId.length - 4);
```
himaniyadav: Is it guaranteed that the postId is always going to be the first part of the likeId? I'm just not familiar with how you're getting the values, but if you're constructing the ids yourself this should be fine

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,193 @@
...
+  commentBox.className = "comment-input";
+  commentBox.id = post.postId + "comment-input";
+  commentBoxDiv.appendChild(commentBox);
+
+  const commentBtn = document.createElement('button');
+  commentBtn.className = "post-btn align-vertical comment-btn";
+  commentBtn.type = "submit";
+  commentBtn.id = post.postId;
+  commentBtn.innerHTML = "<img class='small-icon' src='images/send_plane.png' alt/>";
+  commentBtn.onclick = "postComment()";
```
himaniyadav: Doesn't postComment() take parameters? Does this still work?

himaniyadav (changes requested):
A lot of these comments are pretty minor or just questions on my end!

capstone/src/main/java/com/google/sps/Objects/Comment.java
g-asweet: You don't really want profile pictures on a per-comment basis. If I commented 10 times, you wouldn't want to send 10 copies of the profile picture.

Using a userId here, you could find all relevant comments, then find all users by Id, and just retrieve each profile picture one time.

capstone/src/main/java/com/google/sps/Objects/Post.java
g-asweet: I think this is the same comment --- each user only has one profile picture. Each comment is only made by one user, but you don't want to have a bunch of copies of this user saved everywhere. If the DB handles this and there's ACTUALLY only one user saved, but it's referenced in multiple places then you could have a user saved here.

But from a user ID, you should be able to query the DB and get the user you need to get the profile picture or any other relevant data, right?

capstone/src/main/java/com/google/sps/Objects/Post.java
g-asweet: +1

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  private Post getPostEntity(Entity entity) {
+    long postId = entity.getKey().getId();
+    long timestamp = (long) entity.getProperty("timestamp");
+    String authorId = (String) entity.getProperty("authorId");
+    String postText = (String) entity.getProperty("postText");
+    String challengeName = (String) entity.getProperty("challengeName");
+    String img = (String) entity.getProperty("img");
+    HashSet<String> likes = (entity.getProperty("likes") == null) ? new HashSet<>() : new HashSet<String>((ArrayList<String>) entity.getProperty("likes"));
```
g-asweet: More than 80 chars wide:

```java
HashSet<String> likes = foo
  ? bar
  : ick;
```

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
g-asweet: One of these has a space after `(String)` and one does not. Pick one and make it consistent please.

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/Objects/Post.java
himaniyadav: Yes you should be able to query the DB with the userId but I think that would involve an extra step. If you return the userId with the comment, you would have to: get comment from DB -> return Comment object as JSON -> get user from DB using userId -> return User data as JSON.

Whereas if you stored the needed user data in the comment object, the process would look like this: get comment from DB -> use userId stored in DB to get needed user data -> return Comment object (which contains desired User data) as JSON.

I'm not 100% sure if I'm thinking about this the right way but I don't think the first scenario would necessarily save space because you would still end up with multiple copies of the returned user data

lxcyqx: I think in the first scenario we wouldn't have to store the User object with the Comment in the database, we would just be storing a string for userId. Then we use the userId to get the relevant user information. So I think it does end up saving space in the database since we don't have to save all the User info with each Comment. It's just that now we need to take the userId and query the database each time to get the relevant User information

capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -46,15 +49,19 @@ public String getImg() {
...
+
+  public void removeLike(String userId) {
```
lxcyqx: I think if you remove something that doesn't exist in the set it'll raise an error so you might want to write in a check

capstone/src/main/webapp/group.html
anikabagga: Nope, removed it for the time being so I could work on my features. Will pull from master and add back before pushing!

capstone/src/main/webapp/scripts/post.js
anikabagga: I'm constructing the likeId to be the postId + "like" so I'm using substring to just get the postId.

capstone/src/main/java/com/google/sps/Objects/Post.java
himaniyadav: Oh, I thought we would only be storing the userId in the Database to save space in both scenarios. The difference would be in the first scenario, we would return the userId as part of the Comment object while in the second scenario we would be returning a User/whatever user data we need as part of the Comment object.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
anikabagga: Yep, this will be changed to userId from UserService.

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
lxcyqx, line 60: Since you're checking if the likes ArrayList contains userId, it would be better to convert the ArrayList to a HashSet for constant time lookup

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -0,0 +1,156 @@
...
+
+    // Creates entity with submitted data and add to database
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(createPostEntity(timestamp, authorId, postText, challengeName, img, likes, comments));
+
+    // Redirect back to the HTML page.
+    response.sendRedirect("/group.html");
+  }
+
+  private Entity createPostEntity(long timestamp, String authorId, String postText, String challengeName, String img, ArrayList<String> likes, ArrayList<Comment> comments) {
```
lxcyqx: Over 80 characters

capstone/src/main/java/com/google/sps/servlets/UpdateLikesServlet.java
lxcyqx: You can use this https://www.programcreek.com/java-api-examples/?class=javax.servlet.http.HttpServletResponse&method=sendError and this https://stackoverflow.com/questions/3885725/setting-both-error-message-and-body-from-servlet

capstone/src/main/webapp/scripts/post.js
```diff
@@ -0,0 +1,193 @@
...
+  commentBoxDiv.appendChild(commentBtn);
+
+  return commentBoxDiv;
+}
+
+// Gets URL for uploaded image
+function fetchBlobstoreUrlAndShowForm() {
+  console.log("in fetch");
```
lxcyqx: nit: remove console.log

capstone/src/main/java/com/google/sps/Objects/Post.java
lxcyqx: So if I'm understanding this clearly, the two options are:
1. Comment object with just userId
2. Comment object with things like profile picture, name

In the first scenario we would then need a wrapper object like CommentResponse that takes this Comment object along with other user info and send it to the front end
In the second scenario we would only be sending the comment object back
It doesn't seem like either is taking up more space than the other in terms of database usage, since either way in the database we are just storing userId

anikabagga: For a post, we need the userId, userImage, and userName. For a comment, we need the userId and userImage. As Andrew said above, if a user posted or commented 10 times, we wouldn't want 10 copies of that information. We could loop through all the userId's and just retrieve the user info once?

himaniyadav: Okay, that definitely makes sense

capstone/src/main/webapp/scripts/post.js
anikabagga: forgot to remove

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: Yes, this creates a list of likedPosts by the current user

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: I haven't added the user integration with posts, it's just a placeholder for now. I'll change the variable to authorName to avoid confusion.

capstone/src/main/java/com/google/sps/Objects/Comment.java
anikabagga: Going to leave this as userId so we can avoid pulling the same information multiple times.

capstone/src/main/java/com/google/sps/servlets/BlobstoreServlet.java
anikabagga: Makes sense! I changed the file to PostImageServlet. There is another servlet named BlobstoreKeyServlet that can be used for other functions so I haven't changed that name.

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: This method is going to be used for Blobstore. It gets the key that points to the blobstore image url. The formInputElementName is the html name for the input image element. I replaced "image" with formInputElementName.

capstone/src/main/java/com/google/sps/servlets/UpdateLikesServlet.java
anikabagga: sounds good! I'll make sure to add an errorhandler which displays the kind of error.

capstone/src/main/java/com/google/sps/Objects/Post.java
anikabagga: formatted according to above

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
himaniyadav: got it, makes sense!

capstone/src/main/java/com/google/sps/Objects/Post.java
anikabagga: for HashSets I think it returns false if the item doesn't exist

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -0,0 +1,64 @@
...
+  private EmbeddedEntity createCommentEntity(String commentText, String user) {
+    EmbeddedEntity commentEntity = new EmbeddedEntity();
+    commentEntity.setProperty("timestamp", System.currentTimeMillis());
+    commentEntity.setProperty("commentText", commentText);
+    commentEntity.setProperty("userId", user);
+    return commentEntity;
+  }
+
+  private Entity getPostFromId(HttpServletResponse response, long postId,
+      DatastoreService datastore) throws IOException {
```
g-asweet: https://github.com/google/google-java-format/wiki/The-Rectangle-Rule#the-rectangle-rule

How you've chosen to break this up into multiple lines could be more clear

```
The Rectangle Rule is a big rule that helps to promote readability. Many other possible rules promote readability too, the Rectangle Rule is simple and broad in its implications.
```

It's an arbitrary rule that helps keep a lot consistent, and is already established throughout most of your codebase.

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -21,50 +22,89 @@
...
+    }
+
+    Post userPost = new Post(postId, authorId, postText, comments, challengeName, timestamp, img, likes);
     return userPost;
   }
 
-  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+  private void createCommentObjectList(ArrayList<Comment> comments, Entity entity) {
+    ArrayList<EmbeddedEntity> commentEntitys = (ArrayList<EmbeddedEntity>) entity.getProperty("comments");
+    for(EmbeddedEntity comment: commentEntitys) {
```
g-asweet: `for(` implies it's a function. `for (` better identifies it as a reserved word (if/for/catch)

https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -73,21 +113,50 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     taskEntity.setProperty("likes", likes);
-    taskEntity.setProperty("comments", comments);
+    taskEntity.setProperty("comments", createCommentEntities(comments));
     return taskEntity;
 	}
+
+  // Create list of comment entities from list of comment objects
+  private ArrayList<EmbeddedEntity> createCommentEntities(ArrayList<Comment> comments) {
+    ArrayList<EmbeddedEntity> allComments = new ArrayList<>();
+    for(Comment comment: comments) {
```
g-asweet: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

capstone/src/main/java/com/google/sps/servlets/UpdateLikesServlet.java
```diff
@@ -0,0 +1,70 @@
...
+    this.getUpdatedVotes(isLiked, userId, likes);
+
+    // Update datastore
+    post.setProperty("likes", likes);
+    datastore.put(post);
+  }
+
+  private void getUpdatedVotes(boolean isLiked, String userId, ArrayList<String> likes) {
+    if(isLiked) {
+      if(!likes.contains(userId)) likes.add(userId);
```
g-asweet: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

capstone/src/main/java/com/google/sps/servlets/UpdateLikesServlet.java
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+      likes.remove(userId);
+    }
+  }
+
+  private String getUserId(HttpServletResponse response) throws IOException {
+  UserService userService = UserServiceFactory.getUserService();
+    if(userService.isUserLoggedIn()) {
```
g-asweet: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -0,0 +1,64 @@
...
+    // Get post id and comment text
+    Long postId = Long.parseLong(request.getParameter("id"));
+    String commentText = request.getParameter("comment-text");
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity postEntity = this.getPostFromId(response, postId, datastore);
+    ArrayList<EmbeddedEntity> allComments = (ArrayList<EmbeddedEntity>) postEntity.getProperty("comments");
+
+    // Create comment entity and add to comment arraylist for post
+    if(allComments == null) {
```
g-asweet: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -21,50 +22,89 @@
...
     Query query = new Query("Post").addSort("timestamp", SortDirection.DESCENDING);
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     PreparedQuery results = datastore.prepare(query);
 
     List<Post> posts = new ArrayList<>();
+    List<Long> likedPosts = new ArrayList<>();
     for (Entity entity : results.asIterable()) {
       posts.add(getPostEntity(entity));
+      ArrayList<String> likes = (ArrayList<String>) entity.getProperty("likes");
+      if(likes != null && likes.contains(userId)) {
```
g-asweet: https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -21,50 +22,89 @@
...
+    ArrayList<Comment> comments = new ArrayList<>();
+  
+    // Create list of comment objects from list of comment entities 
+    if (entity.getProperty("comments") != null) {
+      createCommentObjectList(comments, entity);
+    }
+
+    Post userPost = new Post(postId, authorId, postText, comments, challengeName, timestamp, img, likes);
     return userPost;
   }
```
g-asweet: Would it make more sense to define this in the Post.java file as a public `fromEntity(Entity entity)` function?

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,27 +1,85 @@
...
+}
+
+// Performs POST request to add comment to post 
+function postComment(buttonId, commentBoxId) {
+    const commentVal = document.getElementById(commentBoxId).value;
+    let request = new Request(`/post-comment?id=${buttonId}&comment-text=${commentVal}`, { method: "POST" });
+    fetch(request).then(() => {
+      loadPosts();
+    });
+}
```
g-asweet: One of these is 2 spaces, the other is 4 spaces. Please make consistent.

g-asweet (changes requested):
Sorry to go back and forth on this one. Please ping me whenever you've finished this round and I'll approve it immediately :)

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,27 +1,85 @@
 function init() {
-  fetchBlobstoreUrlAndShowForm();
   loadPosts();
   getPollOptions();
-
+  uploadImage();
+  //fetchBlobstoreUrlAndShowForm();
```
himaniyadav: this comment can be deleted if you don't need it anymore

himaniyadav (approved):

capstone/src/main/webapp/scripts/post.js
himaniyadav: oops not sure if I reviewed it correctly but I approved the changes in the sense that the stuff I commented on yesterday seems resolved!

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: I have all my create/post entity functions in the servlet files, so I think I would need to move all those functions to their respective classes to be consistent. Are there any drawbacks to leaving the entity portion in the servlet file?

capstone/src/main/webapp/scripts/post.js
anikabagga: removed

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: nvm will make this change! makes sense to move the post object creation to Post.java.

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: sounds good! added spacing

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
anikabagga: added spacing

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -73,21 +97,50 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+
+  // Create list of comment entities from list of comment objects
+  private ArrayList<EmbeddedEntity> createCommentEntities(ArrayList<Comment> comments) {
+    ArrayList<EmbeddedEntity> allComments = new ArrayList<>();
+    for (Comment comment: comments) {
+      EmbeddedEntity commentEntity = new EmbeddedEntity();
+      commentEntity.setProperty("timestamp", comment.getTimestamp());
+      commentEntity.setProperty("commentText", comment.getCommentText());
+      commentEntity.setProperty("userId", comment.getUser());
+      allComments.add(commentEntity);
```
g-asweet, line 127: I think long-term I'd like to see all entity work inside the Objects directory instead of in the servlets

capstone/src/main/webapp/scripts/post.js
```diff
@@ -48,22 +96,97 @@ function createPostText(post) {
...
+    : `${numLikes} like`;
+  const labelString = numLikes == 0 
+    ? "" 
+    : likesString;
+  likesLabel.innerText = `${labelString}`;
+  likesDiv.appendChild(likesLabel);
+
+  // Set like icon state (filled or only outline)
+  const likeIcon = document.createElement('ion-icon');
+  if(likedPosts.includes(post.postId)) {
```
g-asweet: `if (`

capstone/src/main/webapp/scripts/post.js
```diff
@@ -48,22 +96,97 @@ function createPostText(post) {
...
+  likesDiv.appendChild(likeIcon);
+  return likesDiv;
+}
+
+// Create container for all post comments
+function createCommentsContainer(post) {
+  const commentsContainer = document.createElement('div');
+  commentsContainer.className = "comments-content";
+  const allComments = document.createElement('ul');
+  for(comment of post.comments) {
```
g-asweet: `for (`

capstone/src/main/webapp/scripts/post.js
```diff
@@ -48,22 +96,97 @@ function createPostText(post) {
...
+  const likesLabel = document.createElement('p');
+  likesLabel.className = "likes-label vertical-align";
+  likesLabel.id = post.postId + "likes-label";
+  const numLikes = post["likes"].length;
+  const likesString = numLikes > 1 
+    ? `${numLikes} likes` 
+    : `${numLikes} like`;
+  const labelString = numLikes == 0 
+    ? "" 
+    : likesString;
```
g-asweet: If these fit in one line < 80 chars, they can be put inline. This style is mostly used to avoid going over 80 chars wide.

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,27 +1,75 @@
...
+          likeToggled(this.id, false);
+        } else {
+          likeToggled(this.id, true);
+        }
+      });
+    }
+}
+
+function likeToggled(likeId, liked) {
+  let postId = likeId.substring(0, likeId.length - 4);
```
g-asweet, line 52: I'd still like to see something to avoid this magic number 4, or explain it somehow in the code as a preference with a well named const or some other in-code explanation, or in a comment if there is no clever way to cover that.

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,27 +1,75 @@
...
+    addCommentInputListener();
+    addLikeButtonListener();
+  });
+}
+
+function addCommentInputListener() {
+  let elements = document.getElementsByClassName('post-btn align-vertical comment-btn');
+    for (let i = 0; i < elements.length; i++) {
+      elements[i].addEventListener("click", function() {
+        postComment(this.id, this.id + "comment-input")
```
g-asweet: add a semi-colon to this line

g-asweet (approved):

capstone/src/main/webapp/scripts/post.js
himaniyadav: Not sure if you saw this but I asked the same earlier and she commented above that 
> I'm constructing the likeId to be the postId + "like" so I'm using substring to just get the postId.

So yeah I also think using a const in place of 4 might help clear it up?

anikabagga: added a comment to explain the use of substring/4

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: sure! I'll make those changes in another PR

PR #12, opened 2020-07-01
Weekly challenge

lxcyqx:
- Challenge of the week displayed with poll options
- Connected with database
- General logic: once the due date of current challenge is reached, we delete the top poll option from the database, add that option as challenge into database, set that challenge as new weekly challenge
- fix group styling
Screen Shot 2020-07-06 at 2 52 32 PM


capstone/src/main/java/com/google/sps/Objects/Challenge.java
```diff
@@ -28,11 +29,11 @@ public Badge getBadge() {
...
   }
 
-  public ArrayList<User> usersCompleted() {
+  public ArrayList<String> usersCompleted() {
     return usersCompleted;
   }
 
-  public void addUserCompleted(User user) {
+  public void addUserCompleted(String user) {
     this.usersCompleted.add(user);
```
g-asweet: Could you dedup userIds?

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -0,0 +1,64 @@
...
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Challenge").addSort("dueDate", SortDirection.DESCENDING);
+    PreparedQuery results = datastore.prepare(query);
+    // Get most recent challenge in database
+    Entity entity = results.asIterable().iterator().next();
+    String challengeName = (String) entity.getProperty("name");
+    long dueDate = (long) entity.getProperty("dueDate");
+    ArrayList<String> usersCompleted = (ArrayList<String>) entity.getProperty("usersCompleted");
+    // setting badge as null for now
+    Challenge challenge = new Challenge(challengeName, dueDate, null, usersCompleted);
```
g-asweet: To avoid accessing properties and casting items in place, could I request in `Challenge.java`:

```java
public Challenge fromEntity(Entity) {
  ...
}
```

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String challengeName = request.getParameter("name");
+    LocalDateTime dueDate = this.getDueDate(LocalDateTime.now());
+    long dueDateMillis = Timestamp.valueOf(dueDate).getTime();
+    Entity challengeEntity = new Entity("Challenge");
+    challengeEntity.setProperty("name", challengeName);
+    challengeEntity.setProperty("dueDate", dueDateMillis);
+    challengeEntity.setProperty("usersCompleted", new ArrayList<String>());
```
g-asweet: consider doing this in the `Challenge.java` file

Solution 1:
```java
  Challenge challenge = new Challenge(data, that, i, need);
  datastore.put(challenge.toEntity());
```

or even

Solution 2:
```java
  Challenge challenge = new Challenge(data, that, i, need);
  challenge.save();
```

I think I like the first one because then your classes don't have to be aware of the DB.

But if you wanted to integrate the DB into the class objects, I would recommend doing it with inheritance so you can avoid implementing a lot of duplicate logic.

I think if you do Solution 1, you should consider adding an Interface so all your objects have a `fromEntity(Entity entity)` and `toEntity()` function that they have to implement. Reach out if you have questions about interfaces/implementing this/etc.

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -0,0 +1,81 @@
...
+  // Keeps track of max number of votes an option has
+  private int maxVotes = 0;
+  // Keeps track of the option with the max number of votes
+  private long maxVotesId = 0;
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Option").addSort("timestamp", SortDirection.ASCENDING);
+    ;
```
g-asweet: remove this semicolon

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -0,0 +1,81 @@
...
+ * @author lucyqu
+ *
+ */
+@WebServlet("delete-top-option")
+public class DeletePollOptionServlet extends HttpServlet {
+
+  // Keeps track of max number of votes an option has
+  private int maxVotes = 0;
+  // Keeps track of the option with the max number of votes
+  private long maxVotesId = 0;
```
g-asweet: This concerns me having these as values stored on the servlet itself.

1. `setMaxVotesAndId` does not clear the previous values leaving it prone to issues if you have multiple different polls.
2. `deleteEntity` does not update the `maxVotesId` on call which means calling it multiple times would only delete one item.

I think the simplest solution to these problems is to just return maxVotesId from your current `setMaxVotesAndId`(rename the function), and pass that value into deleteEntity, since you don't seem to need `this.maxVotes` outside of `setMaxVotesAndId`

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
```diff
@@ -78,4 +78,18 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+   * @param response    HttpServletResponse
+   * @param errorString error message
+   * @throws IOException exception thrown when cannot write to file
+   */
+  private void sendError(HttpServletResponse response, String errorString) throws IOException {
+    response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Cannot parse to long.");
+    response.getWriter().print("<html><head><title>Oops an error happened!</title></head>");
+    response.getWriter().print("<body>Something bad happened uh-oh!</body>");
+    response.getWriter().println("</html>");
+  }
```
g-asweet: https://github.com/googleinterns/step100-2020/blob/master/capstone/src/main/java/error/ErrorHandler.java

Can you just use the `ErrorHandler` class you created instead of reimplementing this logic?

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -26,16 +27,49 @@ function getPollOptions() {
...
     });
 }
 
+function updatePoll() {
+  fetch("delete-top-option", { method: "POST" }).then(addChallengeToDb);
+}
+
+function addChallengeToDb() {
+  fetch(`challenge?name=${topChallenge}`, { method: "POST" }).then(
+    getChallenge
```
g-asweet: remove `getChallenge` or finish implementing `addChallengeToDb` :)

g-asweet (changes requested):

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -26,16 +27,49 @@ function getPollOptions() {
...
+function checkWeek() {
+  getChallenge();
+  let now = new Date();
+  //Hard coding due date for now
+  let millisTillDueDate =
+    new Date(now.getFullYear(), now.getMonth(), now.getDate(), 21, 8, 0, 0) -
+    now;
+  if (millisTillDueDate >= 0) {
+    setTimeout(updatePoll, millisTillDueDate);
+  }
```
himaniyadav: I might be misunderstanding here but if millisTillDueDate was negative, shouldn't you manually call updatePoll(), because that indicated the current challenge is outdated?

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
himaniyadav: On a related note, can we make the sendError() method static? So we don't have to instantiate  an ErrorHandler every time we want to send an error message

edit: actually I know @anikabagga has been using ErrorHandler already so if changing it makes things more difficult we can leave it as is

lxcyqx: Actually I think it would make sense to make it static

capstone/src/main/webapp/scripts/poll.js
lxcyqx: yeah I still have to write in code that checks how many weeks have passed and update the new challenge accordingly

capstone/src/main/java/com/google/sps/Objects/Challenge.java
```diff
@@ -2,22 +2,25 @@
...
+      ArrayList<String> usersCompleted) {
+    this.dueDate = dueDate;
     this.challengeName = challengeName;
     this.badge = badge;
     this.usersCompleted = usersCompleted;
   }
 
-  public long getTimestamp() {
-    return timestamp;
+  public long getDueDate() {
```
g-asweet, line 22: I like this better!

capstone/src/main/java/com/google/sps/Objects/Challenge.java
```diff
@@ -28,11 +31,28 @@ public Badge getBadge() {
...
   }
 
-  public ArrayList<User> usersCompleted() {
+  public ArrayList<String> usersCompleted() {
     return usersCompleted;
   }
 
-  public void addUserCompleted(User user) {
-    this.usersCompleted.add(user);
+  public void addUserCompleted(String userId) {
```
g-asweet, line 38: Good choice with `userId` over `user`

capstone/src/main/java/com/google/sps/Objects/Option.java
```diff
@@ -51,6 +54,26 @@ public void addVote(String userId) {
...
+  }
+
+  public static Option fromEntity(Entity entity) {
+    long id = entity.getKey().getId();
+    String text = (String) entity.getProperty("text");
+    List<String> votes = (ArrayList<String>) entity.getProperty("votes");
+    return new Option(id, text, votes);
+  }
+
+  public Entity toEntity() {
```
g-asweet, line 71: Didn't mean to force you to make this an instance method, but I personally think this is cleaner this way :)

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -0,0 +1,58 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Challenge").addSort("dueDate", SortDirection.DESCENDING);
+    PreparedQuery results = datastore.prepare(query);
+    // Get most recent challenge in database
+    Entity entity = results.asIterable().iterator().next();
+    Challenge challenge = Challenge.fromEntity(entity);
+    String json = new Gson().toJson(challenge);
+    response.setContentType("application/json");
+    response.getWriter().println(json);
```
g-asweet: Long-term since this is repeated in EVERY servlet, I'd love to see a way to manage this for all servlets without having to repeat these three lines.

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String challengeName = request.getParameter("name");
+    LocalDateTime dueDate = this.getDueDate(LocalDateTime.now());
+    long dueDateMillis = Timestamp.valueOf(dueDate).getTime();
+    Challenge challenge = new Challenge(challengeName, dueDateMillis, null,
+        new ArrayList<String>());
```
g-asweet: https://github.com/google/google-java-format/wiki/The-Rectangle-Rule#the-rectangle-rule

```java
    Challenge challenge = new Challenge(
      challengeName,
      dueDateMillis,
      null,
      new ArrayList<String>()
    );
```

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -0,0 +1,79 @@
...
+    this.deleteEntity(maxVotedId, results, datastore);
+  }
+
+  /**
+   * Sets the variables maxVotes and maxVotesId by iterating through the entities.
+   *
+   * @param results queried results
+   * @return maxVotesId the id of option with maximum number of votes
+   */
+  private long setMaxVotesAndId(PreparedQuery results) {
```
g-asweet: I think this method is poorly named now that it is not setting an instance value, it doesn't actually "setMaxVotes" any more, but it does `getMaxVotedId`

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -0,0 +1,79 @@
...
+    int maxVotes = 0;
+    for (Entity entity : results.asIterable()) {
+      long id = entity.getKey().getId();
+      List<String> votes = (ArrayList<String>) entity.getProperty("votes");
+      int numVotes;
+      if (votes != null) {
+        numVotes = votes.size();
+      } else {
+        numVotes = 0;
+      }
```
g-asweet: I think this could be a ternary.

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -0,0 +1,79 @@
...
+      if (numVotes > maxVotes) {
+        maxVotes = numVotes;
+        maxVotedId = id;
+      }
```
g-asweet, line 78: OR instead of a ternary, I think this could be combined with the above conditional check since you don't care when numVotes = 0 || votes != null

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @param results   queried results
+   * @param datastore database storing information
+   */
+  private void deleteEntity(long maxVotedId, PreparedQuery results, DatastoreService datastore) {
+    for (Entity optionEntity : results.asIterable()) {
+      if (optionEntity.getKey().getId() == maxVotedId) {
+        datastore.delete(optionEntity.getKey());
+        break;
+      }
```
g-asweet: IDK how important this is but I see optionEntity.getKey() called twice. I imagine in both calls it's O(1), and it doesn't matter, but I have a slight personal preference to only call that method once.

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
```diff
@@ -70,12 +67,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     String text = request.getParameter("text");
-    Entity optionEntity = new Entity("Option");
-    long timestamp = System.currentTimeMillis();
-    optionEntity.setProperty("text", text);
-    optionEntity.setProperty("votes", new ArrayList<String>());
-    optionEntity.setProperty("timestamp", timestamp);
+    Option option = new Option(0, text, new ArrayList<String>());
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(optionEntity);
+    datastore.put(option.toEntity());
```
g-asweet, line 69: 😎

g-asweet (dismissed):

capstone/src/main/webapp/group.html
```diff
@@ -1,67 +1,53 @@
...
+		</div>
+	</template>
 
-    </div>
-  </body>
+	<script>
+		window.onload = getPollOptions();
+	</script>
+	</body>
 </html>
```
nicorod-google: It's possible this is likely due to not having pulled from master recently, but there are lots of changes missing as part of your edit. Importantly, Anika's Post related content is being removed. Also, on lines 49-52 I see you are adding an onload function. I think you can add this to the "init()" function mentioned on line 19 in master

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
nicorod-google: Unresolving because I didn't see any changes

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
nicorod-google: +1

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -26,16 +27,49 @@ function getPollOptions() {
...
+      topChallenge = pollData["options"][0]["text"];
       pollData["options"].forEach(option => {
         renderOptionElement(option, maxVotes);
       });
       return pollData["votedOptions"];
     })
     .then(votedOptions => {
       handleCheck(votedOptions);
+    })
+    .then(checkWeek);
```
nicorod-google: Add the "()" to make it more explicit you're calling the function

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -26,16 +27,49 @@ function getPollOptions() {
...
+  fetch("challenge")
+    .then(response => response.json())
+    .then(challengeData => {
+      const weeklyChallenge = document.getElementById("weekly-challenge");
+      weeklyChallenge.innerText = challengeData["challengeName"];
     });
 }
 
+function updatePoll() {
+  fetch("delete-top-option", { method: "POST" }).then(addChallengeToDb);
```
nicorod-google, line 100: Add the "()" to make it more explicit you're calling the function

capstone/src/main/webapp/scripts/poll.js
nicorod-google: I think it is implemented, rather Lucy should add the "()" to make it more explicit you're calling the function

capstone/src/main/java/com/google/sps/servlets/BlobstoreKeyServlet.java
```diff
@@ -1,20 +1,23 @@
...
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+
 import com.google.appengine.api.blobstore.BlobKey;
 import com.google.appengine.api.blobstore.BlobstoreService;
 import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
-import javax.servlet.annotation.WebServlet;
 
-@WebServlet("serve")
+@WebServlet("blobstore-serve")
```
nicorod-google: It's possible that your PR is just outdated and needs to pull from master, however in case you made these changes I'd double check with Anika to make sure there aren't any other locations you need to update this.

capstone/src/main/java/com/google/sps/servlets/BlobstoreServlet.java
```diff
@@ -1,26 +1,18 @@
...
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-@WebServlet("/post-image-handler")
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+
+@WebServlet("/post-image-blobstore")
```
nicorod-google: Similar to previous comment - It's possible that your PR is just outdated and needs to pull from master, however in case you made these changes I'd double check with Anika to make sure there aren't any other locations you need to update this.

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
nicorod-google: +1

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -26,16 +27,49 @@ function getPollOptions() {
...
       return pollData["votedOptions"];
     })
     .then(votedOptions => {
       handleCheck(votedOptions);
+    })
+    .then(checkWeek);
+}
+
+function checkWeek() {
+  getChallenge();
```
nicorod-google: What's the purpose of getChallenge here? Please correct me if I'm wrong, but unless it's relying on the weeklyChallenge element (which I also don't see it is), I don't see a need for it here?

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -26,16 +27,49 @@ function getPollOptions() {
...
+  let now = new Date();
+  //Hard coding due date for now
+  let millisTillDueDate =
+    new Date(now.getFullYear(), now.getMonth(), now.getDate(), 13, 04, 0, 0) -
+    now;
```
nicorod-google: I know you are just hardcoding the value for now, but can you explain to me the logic for the future? Importantly, where are you going to get the dueDate from? Will it be from the getCallenge() which just currently doesn't return a challenge?

nicorod-google (changes requested):

capstone/src/main/webapp/scripts/poll.js
lxcyqx: I've written the logic in the new pull request. I get the due date from the fetch request in getChallenge, and this due date is passed from getChallenge into checkWeek()

capstone/src/main/webapp/scripts/poll.js
lxcyqx: I changed the code in the new pull request so now checkWeek doesn't have getChallenge

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
lxcyqx: also made these changes in new pull request

capstone/src/main/webapp/scripts/poll.js
nicorod-google: Okie dokie thanks for letting me know!

capstone/src/main/webapp/scripts/poll.js
nicorod-google: Okie dokie!

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
lxcyqx: Made changes in new pull request (mult users)-- created ServletHelper class

nicorod-google (approved):

capstone/src/main/webapp/group.html
lxcyqx: pulled from master, merged in the challenge-mult-users branch, and fixed group styling, so now this branch should be up to date with master.

PR #13, opened 2020-07-01
Blobstore - Post Image Upload 

anikabagga:
- User can post with or without an image
- clicking camera icon opens file upload popout
- Image is sized to fit in width of post (height: auto)
Note: Not sure if we want to mandate image sizes to make them fit a frame. Currently the image can be any height, but fits in the width of the post. We can also look into requiring certain resolutions or forcing a frame based on uploaded image size. 

Screenshot 2020-07-01 at 8 20 06 AM

### DIFF
--- a/capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
@@ -34,7 +34,7 @@ public class CommentsServlet extends HttpServlet {
     ArrayList<EmbeddedEntity> allComments = (ArrayList<EmbeddedEntity>) postEntity.getProperty("comments");
 
     // Create comment entity and add to comment arraylist for post
-    if(allComments == null) {
+    if (allComments == null) {
       ArrayList<EmbeddedEntity> comments = new ArrayList<>();
       comments.add(createCommentEntity(commentText, "user"));
       postEntity.setProperty("comments", comments);
--- a/capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
@@ -14,6 +14,8 @@ import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.datastore.KeyFactory;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
 import java.io.PrintWriter;
 import com.google.sps.Objects.Post;
 import com.google.appengine.api.datastore.EntityNotFoundException;
@@ -24,6 +26,12 @@ public class CommentsServlet extends HttpServlet {
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    String userId = "";
+    UserService userService = UserServiceFactory.getUserService();
+    if (userService.isUserLoggedIn()) {
+      userId = userService.getCurrentUser().getUserId();
+    }
    
     // Get post id and comment text
     Long postId = Long.parseLong(request.getParameter("id"));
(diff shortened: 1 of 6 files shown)


capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -0,0 +1,64 @@
...
+    String commentText = request.getParameter("comment-text");
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity postEntity = this.getPostFromId(response, postId, datastore);
+    ArrayList<EmbeddedEntity> allComments = (ArrayList<EmbeddedEntity>) postEntity.getProperty("comments");
+
+    // Create comment entity and add to comment arraylist for post
+    if(allComments == null) {
+      ArrayList<EmbeddedEntity> comments = new ArrayList<>();
+      comments.add(createCommentEntity(commentText, "user"));
```
lxcyqx: Should you be passing in the userId instead of the string "user"

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,27 +1,80 @@
...
+        postComment(this.id, this.id + "comment-input")
+      });
+    }
+}
+
+function addLikeButtonListener() {
+  let likeBtns = document.getElementsByClassName('like-icon vertical-align');
+    for (let i = 0; i < likeBtns.length; i++) {
+      likeBtns[i].addEventListener("click", function() {
+        let postId = parseInt(this.id.substring(0, this.id.length - 4));
```
lxcyqx: not really sure what 4 is but you can store that in a const and since you use 4 in likeToggle as well

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,27 +1,80 @@
...
+}
+
+function createSinglePost(post, likedPosts) {
   const postDiv = document.createElement('div');
   postDiv.className = "post-div";
   postDiv.appendChild(createProfileImg(post));
   postDiv.append(createAuthor(post));
   postDiv.append(createPostText(post));
-  postDiv.append(createCommentBox());
+  console.log(post.img);
```
lxcyqx: remove console.log()

capstone/src/main/webapp/scripts/post.js
g-asweet: +1

g-asweet: More general comment is that this 4 seems fairly arbitrary and undefined to me. Would a helper function help here? a well named constant? See if you can explain what this 4 means to me in the code

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,27 +1,80 @@
...
+          likeToggled(this.id, false);
+        } else {
+          likeToggled(this.id, true);
+        }
+      });
+    }
+}
+
+function likeToggled(likeId, liked) {
+  let postId = likeId.substring(0, likeId.length - 4);
```
g-asweet: same comment for this 4

capstone/src/main/webapp/scripts/post.js
g-asweet: +1

g-asweet (approved):

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
anikabagga: changed to userId

capstone/src/main/webapp/scripts/post.js
anikabagga: I created a function that gets the postId from the likeId since I do this twice. Couldn't think of a constant name that would make sense. If I should clarify anything else, lmk!

himaniyadav (approved):

capstone/src/main/webapp/scripts/post.js
```diff
@@ -47,15 +46,19 @@ function addLikeButtonListener() {
 }
 
 function likeToggled(likeId, liked) {
-  /* likeId = postId + "like"
-  use substring to just get the postId. */
-  let postId = likeId.substring(0, likeId.length - 4);
+  let postId = getPostIdFromsLikesId(likeId);
```
nicorod-google: nit: Make this a const

capstone/src/main/webapp/scripts/post.js
```diff
@@ -33,9 +34,7 @@ function addLikeButtonListener() {
   let likeBtns = document.getElementsByClassName('like-icon vertical-align');
     for (let i = 0; i < likeBtns.length; i++) {
       likeBtns[i].addEventListener("click", function() {
-        /* likeId = postId + "like"
-        use substring to just get the postId. */
-        let postId = parseInt(this.id.substring(0, this.id.length - 4));
+        let postId = parseInt(getPostIdFromsLikesId(this.id));
```
nicorod-google: nit: Make this a const

nicorod-google (approved):

PR #14, opened 2020-07-01
Basic profile page

himaniyadav:
- Build out profile page that displays basic User information:
- Name, email, phone number, interests are displayed
- Have not added: Challenges or Badges information, "Edit Profile" functionality (clicking on the button does nothing), create or join group buttons, Blobstore functionality for profile picture and group header image
- The group that shows up on the User page is hardcoded. Links to the current default group page.

Screenshot:
image


capstone/src/main/webapp/scripts/profile.js
```diff
@@ -0,0 +1,124 @@
...
+
+  phoneContainer = document.getElementById('phone-container');
+  phoneContainer.innerHTML = user.phoneNumber;
+
+  displayInterests(user.interests);
+  displayProfilePicture(user.profilePic);
+
+  getGroupData(user.groups);
+  
+  displayBadges(user.badges); // maybe change this to getBadgeData?
```
lxcyqx, line 52: I think calling it displayBadges is fine. What do you mean add it to Badge and User class?

himaniyadav: Oh oops, those were comments to myself I forgot to remove. I think similar to how I have a separate getGroupData() function that gets data from the servlet, we might need to do the same thing with Badges. In that case, would have to call getBadgeData() before displayBadges(). Another option, which I think I like better, would be where I would just have the Badge data returned with the User object. Right now I'm not turning the list of badgeIds from database into a list of Badge objects - that's what I meant by adding the functionality to the Badge and User classes. It's currently a TODO in my User fromEntity() method

capstone/src/main/java/com/google/sps/Objects/response/UserGroupResponse.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * Creates and returns a UserGroupResponse object given a group Entity.
+   */
+  public static UserGroupResponse fromEntity(Entity entity) {
+    long groupId = (long) entity.getProperty("groupId");
+    String groupName = (String) entity.getProperty("groupName");
+    String headerImg = (String) entity.getProperty("headerImg");
+    // TODO: fix below code, won't work with Challenges as EmbeddedEntities of Groups
```
nicorod-google, line 36: Is this an issue that will happen when a user has multiple groups?

capstone/src/main/webapp/scripts/profile.js
```diff
@@ -0,0 +1,124 @@
...
+  groupLink.href = "group.html";
+  let groupName = groupElementNode.getElementById('group-name');
+  groupName.innerText = "Group Name";
+  groupsContainer.appendChild(groupElementNode);
+
+  for (group of groups) {
+    let groupElementNode = document.importNode(groupElement.content, true);
+
+    let groupContainer = groupElementNode.querySelector('.group-container');
+    // set group-container div's id to be the groupId
```
nicorod-google: Is this a note for something to do in the future? If so, format it as a TODO

capstone/src/main/webapp/scripts/profile.js
```diff
@@ -0,0 +1,124 @@
...
+
+    let groupName = groupElementNode.getElementById('group-name');
+    groupName.innerText = group.groupName;
+
+    let groupImage = groupElementNode.getElementById('header-img');
+    // set image header based off of group.headerImg url
```
nicorod-google: Same comment here as above

capstone/src/main/java/com/google/sps/servlets/CreateNewUserServlet.java
```diff
@@ -43,28 +45,20 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      String userId = userService.getCurrentUser().getUserId();
+      User user = new User(/* userId */ userId,
+                          /* firstName */ first,
+                          /* lastName */ last, 
+                          /* email */ userService.getCurrentUser().getEmail(),
+                          /* phoneNumber */ phone, 
+                          /* profilePic */ "", 
+                          /* badges */ new LinkedHashSet<Badge>(), 
+                          /* groups */ new LinkedHashSet<Long>(), 
+                          /* interests */ interests);
```
g-asweet:
```java
User user = new User(
  userId,
  firstName,
  lastName, 
  /* email= */ userService.getCurrentUser().getEmail(),
  phoneNumber, 
  /* profilePic= */ "", 
  /* badges= */ new LinkedHashSet<Badge>(), 
  /* groups= */ new LinkedHashSet<Long>(),
  interests
);
```

g-asweet: You don't need a comment for a parameter if the parameter and the variable share the same exact name

capstone/src/main/java/com/google/sps/servlets/UserGroupsServlet.java
```diff
@@ -0,0 +1,93 @@
...
+
+    if (userService.isUserLoggedIn()) {
+      String userId = userService.getCurrentUser().getUserId();
+      DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+      Key entityKey = KeyFactory.createKey("User", userId);
+      try {
+        currentUser = datastore.get(entityKey);
+      } catch (EntityNotFoundException e) {
+        ErrorHandler.sendError(response, "User not found.");
+        return;
```
g-asweet: UserDataServlet still returns null on doGet, whereas here you `return;` early. I don't know which is correct, but I would find a consistent solution for error handling and what gets returned to the server. Have you been able to test these methods with some fake values to see what happens in these error cases?

capstone/src/main/java/com/google/sps/servlets/UserGroupsServlet.java
```diff
@@ -0,0 +1,93 @@
...
+ * Servlet that returns a user's groups data from the Datastore.
+ */
+@WebServlet("/user-groups")
+public class UserGroupsServlet extends HttpServlet {
+
+  /**
+   * Gets User's Group data from the Datastore and returns it.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
nicorod-google, line 57: Could you break this up into a couple of helper methods, perhaps getting the user and then getting a user's groups?

nicorod-google (changes requested):

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/Objects/response/UserGroupResponse.java
himaniyadav: I might be misunderstanding the question. The issue here is that I don't think it will work if I just cast the challenges list to ArrayList<Challenge>. I think Challenges will be stored as embedded entities of groups, like a comment is an embedded entity of a post so I'll need to handle it in a similar way. However, we haven't gotten around to coding with different Group entities yet so this hasn't come up.

nicorod-google: Makes sense, thanks for clearing that up!

capstone/src/main/java/com/google/sps/servlets/UserGroupsServlet.java
himaniyadav: Hmm this led me to do some testing in the error cases to see exactly what happens. The error is correctly thrown and the error message html is displayed at the server url. But our ErrorHandler returns an html response, while the javascript parses the response as json, so in addition to the expected error message sent by the server, there's also "Unexpected token < in JSON". So I think we could modify ErrorHandler to send a json response instead, and generate the error message displayed to the user in the javascript. Or we could have the response.sendRedirect() to the servlet page. Not sure which is the better method.

himaniyadav: Okay, we decided that for the users it would be nice to display an alert message with the error message. This would probably be done on the client side, which raises the question of whether we really need ErrorHandler at all. I think if we return null in these cases then it would be easier to check for the error condition on the client side, but ErrorHandler itself wouldn't serve much purpose.

g-asweet: Well you'd likely still want SOMETHING server side to check for an error and classify the error for the user right? If a user tries to delete a post and the error is that the post does not exist, that's going to require something on the server to check for the existence of the post and then report that error. I would suggest modifying the existing ErrorHandler sendError to to just send a json which has any error you'd want to display to the user.

capstone/src/main/java/com/google/sps/servlets/UserGroupsServlet.java
```diff
@@ -47,47 +47,71 @@
...
+      groups = getGroupsFromIds(groupIds, response);
     } else {
       ErrorHandler.sendError(response, "User not logged in.");
       return;
     }
 
+    // If either of these values are null, an error occured. Do not continue.
+    if (currentUser == null || groups == null) {
+      return;
+    }
```
g-asweet, line 77: Should this not also send an error instead of just returning early?

g-asweet (approved):

capstone/src/main/java/com/google/sps/servlets/UserGroupsServlet.java
nicorod-google: Just to clarify, is the "Unexpected token < in JSON" a JS error or something that is displayed to the user?

If you return null in each of these cases, how would you know what is the proper alert message to display to the user on the clientside? 
I had sent Lucy these links to help her set up the ErrorHandler: 
https://www.programcreek.com/java-api-examples/?class=javax.servlet.http.HttpServletResponse&method=sendError
https://stackoverflow.com/questions/3885725/setting-both-error-message-and-body-from-servlet
The way the ErrorHandler has been implemented, you don't need to do anything on the clientside because it'll just display the error message that was sent. 

While I agree that a proper clientside error handler would be useful, you'd need it to work in conjunction with the server side, something along the lines of: server side sends an error code/message and then the clientside decides what to do with the error it receives. Since we are displaying the error to the user ATM, I don't think we should focus on changing it right now but rather as a stretch feature should we have time later.

capstone/src/main/java/com/google/sps/servlets/UserGroupsServlet.java
himaniyadav: If either of those are null, then I think an error message would've been sent already from one of the other methods so it would be redundant.

nicorod-google (approved):

capstone/src/main/java/com/google/sps/servlets/UserGroupsServlet.java
himaniyadav: The unexpected token in JSON was a JS error. Because the ErrorHandler was sending a response with html in it, but the JS was trying to parse it as JSON.

You're right, the issue with just sending nulls would be that you can't distinguish what the error was caused by, so I guess we would need to keep ErrorHandler. I'm pretty sure the ErrorHandler we're using is implemented based off of the links you sent. However, nothing appears on the clientside (the user doesn't see anything's wrong, just that the page has stopped responding). So I think modifying ErrorHandler to send a JSON might be the best solution. I can focus on that later though if you think it's not a priority for right now.

PR #15, opened 2020-07-01
Challenge mult users

lxcyqx:
- Add challenge due date
- Challenge updates according to due date
- fix styling of group page
Screen Shot 2020-07-06 at 2 52 32 PM


capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -40,14 +47,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String challengeName = request.getParameter("name");
     LocalDateTime dueDate = this.getDueDate(LocalDateTime.now());
     long dueDateMillis = Timestamp.valueOf(dueDate).getTime();
-    Challenge challenge = new Challenge(challengeName, dueDateMillis, null,
-        new ArrayList<String>());
+    Challenge challenge =
+        new Challenge(challengeName, dueDateMillis, null, new ArrayList<String>());
```
himaniyadav: you could include an inline comment before the null parameter mentioning what it is

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -70,8 +76,9 @@ private long setMaxVotesAndId(PreparedQuery results) {
    */
   private void deleteEntity(long maxVotedId, PreparedQuery results, DatastoreService datastore) {
     for (Entity optionEntity : results.asIterable()) {
-      if (optionEntity.getKey().getId() == maxVotedId) {
-        datastore.delete(optionEntity.getKey());
+      Key optionEntityKey = optionEntity.getKey();
+      if (optionEntityKey.getId() == maxVotedId) {
+        datastore.delete(optionEntityKey);
         break;
       }
```
himaniyadav: I think you can avoid iterating through the results by creating a Key object with the desired kind and id.
`Key optionKey = KeyFactory.createKey('Option', maxVotedId)`
Then you'll have to put the `datastore.delete(optionKey)` in a try-catch block because it could throw EntityNotFoundException. But this will be O(1) instead of O(n)

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -25,29 +25,38 @@
...
+    String json = "";
+    // Check if there are challenges in database
+    if (Iterables.size(results.asIterable()) > 0) {
+      // Get most recent challenge in database
+      Entity entity = results.asIterable().iterator().next();
+      Challenge challenge = Challenge.fromEntity(entity);
+      ServletHelper.write(response, challenge, "application/json");
+
+    } else {
+      ServletHelper.write(response, null, "application/json");
```
g-asweet: Can you avoid calling ServletHelper.write twice?

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -34,40 +40,73 @@ function getPollOptions() {
...
+    updatePoll();
+  }
 }
 
+/**
+ * Deletes the top poll option, adding that option as a new challenge to the database.
+ */
 function updatePoll() {
-  fetch("delete-top-option", { method: "POST" }).then(addChallengeToDb);
+  fetch("delete-top-option", { method: "POST" }).then(addChallengeToDb());
```
g-asweet: I think `.then(addChallengeToDb);` was correct, otherwise `addChallengeToDb()` would be called and then the result (`undefined`) is passed into then.

g-asweet (approved):

capstone/src/main/webapp/scripts/poll.js
g-asweet: Alternatively, `.then(() => addChallengeToDb())`, but this is not necessary

lxcyqx: so whenever I call a method in .then should I just exclude the parentheses?

PR #16, opened 2020-07-07
Challenge checkbox

lxcyqx:
Ability to mark checkbox as completed, supports multiple users
Screen Shot 2020-07-07 at 1 55 07 PM


capstone/src/main/java/com/google/sps/Objects/Challenge.java
```diff
@@ -31,27 +33,37 @@ public Badge getBadge() {
...
     this.usersCompleted.add(userId);
   }
 
+  public boolean getIsCompleted(String userId) {
+    if (this.usersCompleted == null) {
+      return false;
+    } else if (this.usersCompleted.contains(userId)) {
+      return true;
+    }
+    return false;
```
himaniyadav: nit: could simplify by saying `return this.usersCompleted.contains(userId);`

lxcyqx: good point!

capstone/src/main/java/com/google/sps/Objects/Challenge.java
```diff
@@ -10,13 +10,15 @@
   private final long dueDate;
   private final Badge badge;
   private final ArrayList<String> usersCompleted;
+  private final long id;
 
-  public Challenge(String challengeName, long dueDate, Badge badge,
-      ArrayList<String> usersCompleted) {
+  public Challenge(
+      String challengeName, long dueDate, Badge badge, ArrayList<String> usersCompleted, long id) {
```
anikabagga, line 30: Could we just store the badgeId since the badge isn't being shown with the challenge?

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -27,14 +32,26 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     Query query = new Query("Challenge").addSort("timestamp", SortDirection.DESCENDING);
     PreparedQuery results = datastore.prepare(query);
+    UserService userService = UserServiceFactory.getUserService();
+    String userId = "";
+    if (userService.isUserLoggedIn()) {
+      userId = userService.getCurrentUser().getUserId();
+    } else {
+      ErrorHandler.sendError(response, "User is not logged in.");
```
g-asweet: Have you tested this manually? I had a brief back and forth with @himaniyadav about whether or not you should return after sending an error, and it looks like in this case you would still be sending a challenge response even though the user isn't logged in.

My hunch is you want to return after sending an error, but make sure that the client-side also handles whatever errors you're sending in a reasonable way.

capstone/src/main/java/com/google/sps/servlets/MarkCheckboxServlet.java
```diff
@@ -21,23 +21,24 @@
 
 import error.ErrorHandler;
 
-@WebServlet("/update-votes")
-public class UpdateVotesServlet extends HttpServlet {
+@WebServlet("/mark-checkbox")
```
g-asweet, line 21: Excellent naming, I know what this does now :)

capstone/src/main/java/com/google/sps/servlets/MarkCheckboxServlet.java
```diff
@@ -21,23 +21,24 @@
...
-    Entity optionEntity = this.getEntityFromId(response, optionId, datastore);
-    Set<String> votesSet = this.getUpdatedVotes(optionEntity, isOptionChecked, userId);
+    Entity entity = this.getEntityFromId(response, id, datastore, type);
+    Set<String> votesSet = this.getUpdatedVotes(entity, isChecked, userId);
 
     // Update datastore
-    optionEntity.setProperty("votes", votesSet);
-    datastore.put(optionEntity);
+    entity.setProperty("votes", votesSet);
+    datastore.put(entity);
```
g-asweet, line 41: I don't see any error handling in the above lines of code, and I do see that any time you call doPost, that you're updating the database.

I would think about this more defensively: If somebody wanted to attack this endpoint, what could they send? How do I invalidate an 'attack'?

g-asweet: Some portion of this is directed more broadly:

@anikabagga  @himaniyadav the three of you may want to create an abstract class AuthenticatedServlet which would override doPost and doGet and would check things like:
- a user is logged in and has a valid ID
- (Any other checks you would want for ALL gets/posts)

then you could have something like:

```java
class AuthenticatedServlet extends HttpServlet {
  ...
}

class MarkCheckboxServlet extends AuthenticatedServlet {
  @Override
  public void doPost(String userId, HttpServletRequest request, HttpServletResponse response) throws IOException {
    ...
  }
}
```

Where you can guarantee that the logic you're writing is only called when a valid userId was found

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -74,6 +80,9 @@ function getChallenge() {
     .then(() => checkWeek(dueDateMillis));
```
g-asweet: here you're calling `checkWeek(dueDateMillis)` regardless of `if (challengeData)` was true, meaning `dueDateMillis` may still be 0. Is this intended?

If not, consider moving checkWeek into `if (challengeData)`

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -213,7 +222,19 @@ function renderOptionElement(option, maxVotes) {
...
+  }).then(setTimeout(getPollOptions, TRANSITION_MILLIS));
+}
+
+/**
+ * Handles challenge checkbox change.
+ * @param {String} id
+ * @param {String} checked
+ */
+function markChallenge(id, checked) {
+  console.log("in mark challenge");
```
g-asweet: Can you remove this `console.log`?

g-asweet (dismissed):

capstone/src/main/java/com/google/sps/Objects/response/ChallengeResponse.java
```diff
@@ -0,0 +1,22 @@
...
+import com.google.sps.Objects.Challenge;
+
+public class ChallengeResponse {
+
+  private Challenge challenge;
+  private boolean isCompleted;
+
+  public ChallengeResponse(Challenge challenge, boolean isCompleted) {
+    this.setChallenge(challenge);
+    this.isCompleted = isCompleted;
```
nicorod-google: Could you rename this, and elsewhere in the Challenge class and Challenge Servlet, to something like userHasCompleted instead? I feel isCompleted is a little vague as to its intent, and I needed to look through the code to identify what exactly the purpose of this was

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/Challenge.java
lxcyqx: Aren't we showing the badge icon with the weekly challenge?

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
lxcyqx: Yeah I also noticed that yesterday. Anika, Himani and I discussed this and I think we are going to send an alert on the client side and just get rid of ErrorHandler

capstone/src/main/java/com/google/sps/Objects/response/ChallengeResponse.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * This class is used in the ChallengeServlet to wrap together the current challenge and whether the
+ * current user has completed the challenge.
+ *
+ * @author lucyqu
+ */
+public class ChallengeResponse {
+
+  private Challenge challenge;
+  private boolean isCompleted;
```
nicorod-google: Please also change all of the isCompleteds here to hasUserCompleted

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -61,22 +61,50 @@ function getChallenge() {
...
+      )[0];
+      const challengeLabel = document.getElementById("challenge-label");
       if (challengeData) {
-        weeklyChallenge.innerText = challengeData["challengeName"];
-        dueDateMillis = challengeData["dueDate"];
+        setChallengeCheckboxVisibility("visible");
+        const id = challengeData["challenge"]["id"];
+        challengeCheckbox.id = id;
+        challengeLabel.setAttribute("for", id);
+        challengeCheckbox.checked = challengeData["isCompleted"];
```
nicorod-google: This would need to be hasUserCompleted after changing the ChallengeResponse

capstone/src/main/java/com/google/sps/Objects/response/ChallengeResponse.java
lxcyqx: I tried doing this and went back and forth several times but for whatever reason when I change it to hasUserCompleted, the checkbox status doesn't save when the page is reloaded. Not sure what the issue is

capstone/src/main/webapp/group.html
```diff
@@ -1,101 +1,94 @@
...
+		</div>
+	</template>
 
+	<script>
+		window.onload = getPollOptions();
+	</script>
+</body>
 
-  
+</html>
```
nicorod-google: I think this is largely my fault for not having noticed the preference in the cloud shell, but I'd recommend changing the tabSize from 4 to 2 (File -> Settings -> Open Preferences -> editor -> tabSize). There looks to be a lot of unnecessary formatting change here, I think mostly coming from that

nicorod-google (changes requested):

capstone/src/main/webapp/group.html
lxcyqx: hmm the tab size was already set to 2 i don't know why that formatting happened

capstone/src/main/java/com/google/sps/Objects/response/ChallengeResponse.java
```diff
@@ -0,0 +1,26 @@
...
+
+  /**
+   * Constructor to set the instance variables.
+   *
+   * @param challenge Challenge object
+   * @param isCompleted boolean indicating whether user has completed challenge
+   */
+  public ChallengeResponse(Challenge challenge, boolean userHasCompleted) {
+    this.challenge = challenge;
+    this.userHasCompleted = userHasCompleted;
```
nicorod-google: Sorry for being nitpicky, but boolean variables should start with is/has. I don't think you can see this g3doc so here are the guidelines:
By convention, boolean variables are usually named "foo" instead of "isFoo" so that it's a noun phrase rather than verb phrase just like most other variables. On the other hand, methods that return boolean are usually named "isFoo" so that it's a verb phrase just like most other methods.

I know you follow the noun phrase but we also want them to start the same

```suggestion
  private boolean hasUserCompleted;

  /**
   * Constructor to set the instance variables.
   *
   * @param challenge Challenge object
   * @param hasUserCompleted boolean indicating whether user has completed challenge
   */
  public ChallengeResponse(Challenge challenge, boolean hasUserCompleted) {
    this.challenge = challenge;
    this.hasUserCompleted = hasUserCompleted;
```

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -61,22 +61,50 @@ function getChallenge() {
...
+      )[0];
+      const challengeLabel = document.getElementById("challenge-label");
       if (challengeData) {
-        weeklyChallenge.innerText = challengeData["challengeName"];
-        dueDateMillis = challengeData["dueDate"];
+        setChallengeCheckboxVisibility("visible");
+        const id = challengeData["challenge"]["id"];
+        challengeCheckbox.id = id;
+        challengeLabel.setAttribute("for", id);
+        challengeCheckbox.checked = challengeData["userHasCompleted"];
```
nicorod-google:
```suggestion
        challengeCheckbox.checked = challengeData["hasUserCompleted"];
```

capstone/src/main/java/com/google/sps/servlets/MarkCheckboxServlet.java
```diff
@@ -0,0 +1,128 @@
...
+  @Override
+  public void doPost(String userId, HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+    String idString = request.getParameter("id");
+    boolean isChecked = Boolean.parseBoolean(request.getParameter("checked"));
+    // type representing whether checkbox is for Option or for Challenge
+    String type = request.getParameter("type");
+    long id = this.parseToLong(response, idString);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    //    String userId = this.getUserId(response);
```
nicorod-google: nit: remove this line

nicorod-google (changes requested):

nicorod-google (approved):

PR #17, opened 2020-07-07
User and authentication tests

himaniyadav:
Add tests for the following servlets:
- CreateNewUserServlet.java
- CheckNewUserServlet.java
- LoginServlet.java

All tests run successfully.


capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
```diff
@@ -0,0 +1,136 @@
...
+    // Set up a fake HTTP response.
+    responseWriter = new StringWriter();
+    when(mockResponse.getWriter()).thenReturn(new PrintWriter(responseWriter));
+
+    checkNewUserServlet = new CheckNewUserServlet();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
```
lxcyqx, line 97: You should set your other instance variables to null (responseWriter, datastore, checkNewUserServlet)

capstone/src/test/java/com/google/sps/servlets/LoginServletTest.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @Test
+  public void doGet_userIsLoggedOut() throws Exception {
+    helper.setEnvIsLoggedIn(false);
+
+    loginServlet.doGet(mockRequest, mockResponse);
+    String response = responseWriter.toString();
+
+    assertThat(response).contains("false");
```
lxcyqx, line 92: for assertThat shouldn't there be a second field 
`assertThat(current().containsKey("1"), is(true));`

capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -98,6 +98,26 @@ public void addGroup(long newGroupId) {
...
+    User user = (User) other;
+    return userId.equals(user.userId) &&
+        firstName.equals(user.firstName) &&
+        lastName.equals(user.lastName) &&
+        email.equals(user.email) &&
+        phoneNumber.equals(user.phoneNumber) &&
+        profilePic.equals(user.profilePic) &&
+        interests.containsAll(user.interests) && user.interests.containsAll(interests) &&
+        groups.containsAll(user.groups) && user.groups.containsAll(groups) &&
+        badges.containsAll(user.badges) && user.badges.containsAll(badges);
```
g-asweet, line 118: I like this, but I'm concerned about somebody adding a new field to the User object and not updating the equals call accidentally...

I'm wondering if you can abuse any other logic that you KNOW you have to update. `toEntity/fromEntity` or `new Gson().toJson` and a string comparison?

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.mockito.MockitoAnnotations;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.ArrayList;
+
+/**
+ * Unit tests for {@link CheckNewUserServlet}.
+ */
+ @RunWith(JUnit4.class)
+public class CheckNewUserServletTest {
```
g-asweet, line 38: I would also add tests for `toEntity`/`fromEntity`, or at least one round-trip test (`thing == Thing.fromEntity(thing.toEntity())`) that ensures no data is lost i.e. when a new field is added to the object, this round-trip test would make sure that the developer added it to `toEntity` and `fromEntity` and not on or the other.

This test would still pass if a user didn't add the new field to EITHER, so I would see if you can also come up with a testing strategy that would cause a test to fail if a user added a new field to the object but didn't add it to `toEntity`/`fromEntity`

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
```diff
@@ -0,0 +1,136 @@
...
+  private static final String CURRENT_USER_EMAIL = "test@mctest.com";
+  private static final String CURRENT_USER_ID = "testy-mc-test";
```
g-asweet, line 40: I think officially it would be `testy-mc-test-face` :P

capstone/src/test/java/com/google/sps/servlets/LoginServletTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Mock private HttpServletRequest mockRequest;
+  @Mock private HttpServletResponse mockResponse;
+  private StringWriter responseWriter;
+  private DatastoreService datastore;
+  private LoginServlet loginServlet;
+
+  @Before
+  public void setUp() throws Exception {
+    MockitoAnnotations.initMocks(this);
```
g-asweet, line 62: This line is crucial because it causes all the mocks to be reset before each test.

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
g-asweet: I'm not sure I follow, do you just mean instead of

`private StringWriter responseWriter;`

she should have

`private StringWriter responseWriter = null;`?

I actually think the approach of **not** setting it null makes the tests fail much more likely if a new variable was added and you forgot to set it up for a test in `setUp` or `tearDown`, which I would argue is a good thing that reminds the developer to add a new variable to the `setUp`/`tearDown`

The only note I would have is to make sure that any variable that you're declaring up above should either be final or explicitly updated in `setUp` and/or `tearDown` to ensure you don't accidentally leak data between tests, which you appear to be doing.

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+  @Test
+  public void doPost_newUserVisiting() throws Exception {
+    checkNewUserServlet.doGet(mockRequest, mockResponse);
+    String response = responseWriter.toString();
+    assertThat(response).contains(/* isUserNew= */ "true");
+  }
+
+  @Test
+  public void doPost_existingUserVisiting() throws Exception {
```
g-asweet: I think these two should be named `doGet_` instead of `doPost_` right?

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
```diff
@@ -0,0 +1,136 @@
+package com.google.sps.servlets;
```
g-asweet, line 1: Sorry I can't comment on a file you're not editing in the pull request. This also may be fully out of the scope of this pull request, but I have a question. Looking at CheckNewUserServlet:

```java
    boolean isUserNew = false;

    if (userService.isUserLoggedIn()) {
      String userId = userService.getCurrentUser().getUserId();

      // Check if user already exists in Datastore. If so, do nothing.
      Key entityKey = KeyFactory.createKey("User", userId);
      try {
        datastore.get(entityKey);
      } catch (EntityNotFoundException e) {
        // If the user doesn't exist in Datastore.
        isUserNew = true;
      }
    }

    String json = String.format("{\"isUserNew\": %b}", isUserNew);
```

If `userService.isUserLoggedIn()` is false, you return a normal response where you claim `{isUserNew: false}` but you would have no way of knowing since `isUserLoggedIn()` returned false, right? I may be missing something, but should you error out if the userService can't identify the user? Identify them as a new user? Redirect them to a different page?

capstone/src/test/java/com/google/sps/servlets/CreateNewUserServletTest.java
```diff
@@ -0,0 +1,124 @@
...
+    when(mockRequest.getParameter("last")).thenReturn(USER_1.getLastName());
+    when(mockRequest.getParameter("phone")).thenReturn(USER_1.getPhoneNumber());
+    when(mockRequest.getParameter("interests")).thenReturn("Testing, Dancing");
+
+    createNewUserServlet.doPost(mockRequest, mockResponse);
+
+    Key userKey = KeyFactory.createKey("User", USER_ID);
+    datastore.get(userKey); 
+
+    helper.setEnvIsLoggedIn(true);
```
g-asweet: There don't appear to be any assertions in this test. If this is intended, I would add a comment after the line of code that should trigger the exception, and ideally the test would end at that comment, since you should try to just test the exception, any other testing should be done in another test.

g-asweet: Do you need to call `helper.setEnvIsLoggedIn(true)` in order to trigger the exception?

capstone/src/test/java/com/google/sps/servlets/LoginServletTest.java
```diff
@@ -0,0 +1,97 @@
...
+  public void doGet_userIsLoggedOut() throws Exception {
+    helper.setEnvIsLoggedIn(false);
+
+    loginServlet.doGet(mockRequest, mockResponse);
+    String response = responseWriter.toString();
+
+    assertThat(response).contains("false");
+    assertFalse(response.contains(USER_EMAIL));
+
+    helper.setEnvIsLoggedIn(true);
```
g-asweet: Is this line necessary? You seem to be mocking state after the test, and not testing it at all. And if you were to add additional assertions after this line, would you be testing the servlet, or the helper at that point?

g-asweet: If you're trying to restore state after the test, I think this should already be handled in `helper.setUp()` and `helper.tearDown()`

and if not (please confirm), then I would DEFINITELY move this line into `setUp` because attempting to restore state in a test is something which is not guaranteed. Assume a test can crash and fail at any point, whereas you can safely assume setUp is guaranteed to run before a test runs.

g-asweet (dismissed):

capstone/src/test/java/com/google/sps/servlets/CreateNewUserServletTest.java
himaniyadav: No, but I thought the tests should be independent of one another, and since I modified the initial value (logged in = true), I thought I should change it back to what it was before the test.

capstone/src/test/java/com/google/sps/servlets/LoginServletTest.java
himaniyadav: Yeah, I was trying to restore it's state after the test. I didn't realize that setUp() and tearDown() were run before and after each test. Will fix!

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
himaniyadav: That's true, I think it should send an error. Because ideally /checknewuserservlet is only called after /loginservlet returns true, so the user should never be at this point and logged out.

Also I noticed you mentioned creating an abstract class AuthenticatedServlet that verifies authentication in Lucy's PR - if we did that, I guess I wouldn't need the additional check at all?

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
himaniyadav: Oops, yes!

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+  @Test
+  public void doPost_existingUserVisiting() throws Exception {
+    addUserToDatastore(datastore, CURRENT_USER);
+
+    checkNewUserServlet.doGet(mockRequest, mockResponse);
+    String response = responseWriter.toString();
+    assertThat(response).contains(/* isUserNew= */ "false");
+
+    removeUserFromDatastore(datastore, CURRENT_USER);
```
nicorod-google: You shouldn't need to remove the user that was added to the DB at the end of the test because the helper.tearDown() should do that for you. Also, any post processing should be done in the tearDown function instead of after the asserts

nicorod-google (changes requested):

capstone/src/test/java/com/google/sps/servlets/CreateNewUserServletTest.java
himaniyadav: Got it, will do

himaniyadav: Okay got it, will move that to `helper.tearDown()`

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
lxcyqx: I meant that in tearDown she should have `responseWriter = null;`

capstone/src/main/java/com/google/sps/Objects/User.java
himaniyadav: Updated to use fromEntity and Gson. Should I keep the .equals() method in User or no?

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
himaniyadav: Okay. Can tests for toEntity and fromEntity go in CheckNewUserServletTest.java or should they go in a new class specifically to test methods in User.java?

capstone/src/test/java/com/google/sps/servlets/CreateNewUserServletTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+    createNewUserServlet.doPost(mockRequest, mockResponse);
+    Key userKey = KeyFactory.createKey("User", USER_ID);
+    Entity user = datastore.get(userKey);
+
+    String jsonDs = new Gson().toJson(User.fromEntity(user));
+    String jsonCurrent = new Gson().toJson(USER_1);
+    
+    assertTrue(jsonDs.equals(jsonCurrent));
+    //assertTrue(USER_1.equals(User.fromEntity(user)));
```
nicorod-google: nit: remove this line

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
nicorod-google: Since the toEntity and fromEntity methods are specific to the User class, they should be tested independently in a UserTest

capstone/src/test/java/com/google/sps/Objects/UserTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Unit tests for {@link User}.
+ */
+ @RunWith(JUnit4.class)
+public class UserTest {
+  private static final String USER_EMAIL = "test@mctest.com";
+  private static final String USER_ID = "testy-mc-test";
+
+  // https://cloud.google.com/appengine/docs/java/tools/localunittesting
```
nicorod-google: You probably don't need this line

capstone/src/test/java/com/google/sps/Objects/UserTest.java
```diff
@@ -0,0 +1,88 @@
...
+                      "com.google.appengine.api.users.UserService.user_id_key", USER_ID)));
+
+  private static final ArrayList<String> INTERESTS_LIST = new ArrayList<String>( 
+      Arrays.asList("Testing", "Dancing"));
+  private static final User USER_1 = new User(USER_ID, "Test", "McTest", USER_EMAIL, 
+                          /* phoneNumber= */ "123-456-7890", 
+                          /* profilePic= */ "", 
+                          /* badges= */ new LinkedHashSet<Badge>(), 
+                          /* groups= */ new LinkedHashSet<Long>(), 
+                          /* interests= */ INTERESTS_LIST);
```
nicorod-google: All of the argument should be on a different line if it's over 80 characters
```suggestion
  private static final User USER_1 = new User(
                          USER_ID, 
                          "Test", 
                          "McTest", 
                          USER_EMAIL, 
                          /* phoneNumber= */ "123-456-7890", 
                          /* profilePic= */ "", 
                          /* badges= */ new LinkedHashSet<Badge>(), 
                          /* groups= */ new LinkedHashSet<Long>(), 
                          /* interests= */ INTERESTS_LIST);
```

nicorod-google (approved):

capstone/src/test/java/com/google/sps/servlets/CheckNewUserServletTest.java
himaniyadav: > This test would still pass if a user didn't add the new field to EITHER, so I would see if you can also come up with a testing strategy that would cause a test to fail if a user added a new field to the object but didn't add it to `toEntity/fromEntity`

In fromEntity, a User object is made using the constructor. If we added a field to the object and added that field to the constructor, but didn't update the fromEntity method, then fromEntity would throw an error. This would force fromEntity to be updated when the object is updated, which in turn would cause the round-trip test to fail if toEntity wasn't also updated.

PR #18, opened 2020-07-08
Group Members 

anikabagga:
- Created group member scrollable panel 
- Clicking on a member profile picture opens a detailed member profile modal 
- Made two new servlets (AllGroupMembersServlet & GroupMemberServlet)
- AllGroupMembersServlet: Retrieves basic information (name & profile img) of all members in group to display
- GroupMemberServlet: Provides more detailed information for a userId (name, profile img, badges) and can add a new member to group 

Note: I haven't implemented the badges panel for a member profile since this overlaps with the user profile. 

image


capstone/src/main/java/com/google/sps/Objects/response/BasicMemberResponse.java
```diff
@@ -0,0 +1,79 @@
...
+  
+  /**
+   * Constructs a BasicMemberResponse object from an entity
+   *
+   * @param profilePic String url of profile image 
+   * @param firstName String of user first name 
+   * @param lastName String of user last name  
+   * @param userId user id 
+   */
+  public static BasicMemberResponse fromEntity(Entity entity) {
```
g-asweet: I haven't played around with this and you guys have, but I have a question: Seeing as `BasicMemberResponse` appears to be the same as `MemberResponse` minus the badges, if you don't set any badges, can you avoid sending `badges: null/undefined/[]` while using a MemberResponse instead of a BasicMemberResponse?

I'd hate to see you have to create a new version of the exact same object every time you don't want to include a field.

I would still maybe even suggest using MemberResponse with `badges: null/undefined/[]/[actual, badges, here]` instead of having to create a new BasicMemberResponse when the two are so similar. Thoughts?

capstone/src/main/java/com/google/sps/servlets/AllGroupMembersServlet.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EntityNotFoundException;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;;
```
g-asweet: `import java.util.List;;` -> `import java.util.List;`

capstone/src/main/webapp/styles/group.css
```diff
@@ -279,17 +288,17 @@ p {
   height: 60px;
   width: 60px;
   border-radius: 50%;
-  background-color: #f5ceda;
+  background-color: #F5CEDA;;
```
g-asweet: `#F5CEDA;;` -> `#F5CEDA;`

capstone/src/main/java/com/google/sps/Objects/response/MemberResponse.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Constructs a MemberResponse object from an entity
+   *
+   * @param profilePic String url of profile image 
+   * @param firstName String of user first name 
+   * @param lastName String of user last name  
+   * @param badges List of badge objects 
+   * @param userId user id 
+   */
+  public static MemberResponse fromEntity(Entity entity) {
```
g-asweet: If you want to avoid adding badges in some scenarios, you could do the following:

```java
  public static MemberResponse fromEntity(Entity entity) {
    return MemberResponse.fromEntity(entity, /* includeBadges= */ true);
  }

  public static MemberResponse fromEntity(Entity entity, boolean includeBadges) {
    [Insert Your Current Logic]
    [check if includeBadges before doing anything with badges]
  }
```

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -0,0 +1,92 @@
...
+  private String getUserId(HttpServletResponse response) throws IOException {
+  UserService userService = UserServiceFactory.getUserService();
+    if (userService.isUserLoggedIn()) {
+      return userService.getCurrentUser().getUserId();
+    }
+    ErrorHandler.sendError(response, "User is not logged in.");
+    return "";
+  }
+
+  private Entity getGroupFromId(HttpServletResponse response, long groupId, DatastoreService datastore) throws IOException {
```
g-asweet: over 80 chars long

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -0,0 +1,92 @@
...
+    try {
+      return datastore.get(KeyFactory.createKey("Group", groupId));
+    } catch (EntityNotFoundException e) {
+      ErrorHandler.sendError(response, "Group does not exist.");
+      return null;
+    }
+  }
+
+  private Entity getUserFromId(HttpServletResponse response, String userId, DatastoreService datastore) throws IOException {
```
g-asweet: over 80 chars long

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  // Adds a new member to a group 
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    
+    String userId = this.getUserId(response);
+    Long groupId = Long.parseLong(request.getParameter("groupId"));
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity group = this.getGroupFromId(response, groupId, datastore);
```
g-asweet, line 34: `getGroupFromId` claims to throw IOException, but it also catches it which means it will not throw the exception. Instead, it returns null. Then, you call `group.getProperty` without a null check which could cause this to crash.

Consider making `getGroupFromId` actually throw an error, in which case any time you call it you'd need to do a `try`/`catch`, or keep what you have, remove the `throws IOException` from getGroupFromId, and just get that the group isn't null before you use it.

capstone/src/main/webapp/scripts/members.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function createMemberProfileImg(memResponse){
+  const memberProfileImgDiv = document.createElement('div');
+  // If user doesn't have image, replace with blank large yellow profile circle 
+  if (memResponse.profilePic == null || memResponse.profilePic == "") {
+    memberProfileImgDiv.className = "member-img-blank large-member-img";
+  } else {
+    memberProfileImgDiv.className = "member-img-div large-member-img";
+    let memberProfileImg = document.createElement('img');
```
g-asweet: `const`

capstone/src/main/webapp/scripts/members.js
```diff
@@ -0,0 +1,103 @@
...
+function createMemberComponents(memberInfo) {
+  const memberImgDiv = document.createElement('div');
+  memberImgDiv.id = memberInfo.userId;
+  memberImgDiv.title = memberInfo.firstName + " " + memberInfo.lastName;
+  // If user doesn't have image, replace with blank small yellow profile circle 
+  if (memberInfo.profilePic == null || memberInfo.profilePic == "") {
+    memberImgDiv.className = "member-grid-item member-img-blank small-member-img";
+  } else {
+    memberImgDiv.className = "member-grid-item member-img-div small-member-img";
+    let memberImg = document.createElement('img');
```
g-asweet: `const memberImg`

capstone/src/main/webapp/scripts/members.js
```diff
@@ -0,0 +1,103 @@
...
+    memberImg.className = "member-img small-member-img";
+    memberImg.src = "serve?blob-key=" + memberInfo.profilePic;
+    memberImgDiv.append(memberImg);
+  }
+  return memberImgDiv;
+}
+
+// When group member clicked, show detailed member profile modal
+function addMemberProfileListener() {
+  let elements = document.getElementsByClassName('member-grid-item');
```
g-asweet: `const elements`

g-asweet (changes requested):

capstone/src/main/webapp/styles/members.css
```diff
@@ -0,0 +1,123 @@
...
+  color:  #C4C4C4;
+  height: 280px;
+  padding: 15px;
+  width: 200px;
+  float: right;
+  margin-right: 15%;
+  margin-top: 200px;
+}
+
+#members-container h4 {
```
himaniyadav, line 13: This is just an aesthetic thing haha but could you play around with the text and the plus sign so that they're aligned horizontally with the icons?

capstone/src/main/webapp/styles/group.css
```diff
@@ -299,17 +308,19 @@ p {
...
 #comment-input {
-  background-color: #eeeeee;
+  background-color: #EEEEEE;
   border-radius: 10px;
   align-content: left;
   margin-left: 100px;
   height: 40px;
   width: 70%;
-  color: #8a8686;
+  color: #8A8686;
```
himaniyadav: since we've set up variables for these colors I think it would make sense to use those? `var(--placeholder-gray)` and `var(--text-gray)`

capstone/src/main/webapp/styles/members.css
```diff
@@ -0,0 +1,123 @@
+#members-container {
+  background-color: #fefefe;
+  border-radius: 10%;
+  color:  #C4C4C4;
```
himaniyadav: same note about using the variables: `var(--near-white)` and `var(--badges-gray)`

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -0,0 +1,92 @@
...
+
+@WebServlet("/group-member")
+
+public class GroupMemberServlet extends HttpServlet {
+
+  @Override
+  // Adds a new member to a group 
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    
+    String userId = this.getUserId(response);
```
himaniyadav, line 30: To clarify, this adds the current user to a specified group.. so eventually would this be called when the user enters a code to join a group?

anikabagga: yeah this adds a user to a group for later on

capstone/src/main/java/com/google/sps/Objects/response/BasicMemberResponse.java
lxcyqx: I think sending null for badges and getting rid of BasicMemberResponse makes sense. Why do you say "can you avoid sending badges: null/undefined/[] while using a MemberResponse"

anikabagga: I was trying to avoid sending unnecessary info, but it makes sense to remove BasicMemberResponse and set null for Badges.

capstone/src/main/java/com/google/sps/Objects/response/MemberResponse.java
anikabagga: removed basicMember object and added includeBadges check

capstone/src/main/webapp/styles/members.css
anikabagga: aligned it with the post a status div

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
anikabagga: makes sense, I removed throws IOException and added a check to make sure group isn't null

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -0,0 +1,99 @@
...
+import com.google.gson.Gson;
+import com.google.sps.Objects.response.MemberResponse;
+import error.ErrorHandler;
+
+@WebServlet("/group-member")
+
+public class GroupMemberServlet extends HttpServlet {
+
+  @Override
+  // Adds a new member to a group
```
nicorod-google, line 27: I think I may be a bit confused here, are you adding this just for the future to be able to add a user to a group? Correct me if I'm wrong, but I don't think this is being used at the moment?

capstone/src/main/webapp/scripts/members.js
```diff
@@ -0,0 +1,103 @@
...
+  fetch(`/all-group-members`).then(response => response.json()).then((allGroupMembers) => {
+    const allMembers = allGroupMembers;
+    const memberGrid = document.getElementsByClassName("member-grid-container")[0];
+    memberGrid.innerHTML = '';
+    for (let i = 0; i < allMembers.length; i++) {
+      memberGrid.appendChild(createMemberComponents(allMembers[i]));
+    }
+  }).then(() => {
+    addMemberProfileListener();
+  });
```
nicorod-google: The previous then is not returning a promise, so I think you can just call the addMemberProfileListener() after the for loop

capstone/src/main/webapp/scripts/members.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function createMemberComponents(memberInfo) {
+  const memberImgDiv = document.createElement('div');
+  memberImgDiv.id = memberInfo.userId;
+  memberImgDiv.title = memberInfo.firstName + " " + memberInfo.lastName;
+  // If user doesn't have image, replace with blank small yellow profile circle
```
nicorod-google, line 17: I like this!

capstone/src/main/webapp/scripts/members.js
```diff
@@ -0,0 +1,103 @@
...
+// Show group member profile (profile pic, name, and badges)
+function showMemberProfile(userId){
+  fetch(`/group-member?id=${userId}`).then(response => response.json()).then((memResponse) => {
+    const modalContent = document.getElementsByClassName("modal-content")[0];
+    modalContent.innerHTML = "<span class='close'>&times;</span>";
+    addModalListeners();
+    modalContent.appendChild(createMemberModal(memResponse));
+  }).then(() => {
+    const modal = document.getElementById("memberProfile");
+    modal.style.display = "block";
```
nicorod-google, line 48: Same as the comment above, since the previous then is not returning a promise, you should just put this after the modalContent.append....
If you are using the then's to separate content, you can break it up into functions instead

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
anikabagga: I'm not using it currently, but created it for future use when we need to add a user to a group.

capstone/src/main/webapp/scripts/members.js
anikabagga: I was mainly using it as a separator, but that makes sense. fixed!

capstone/src/main/webapp/scripts/members.js
anikabagga: moved to after modalContent.append

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
anikabagga: I think I have to include the throws IOException since ErrorHandler.sendError also throws IOException.

nicorod-google: This comment applies to the getGroupFromId/getUserFromId method below, not the doPost one here. Andrew is saying that when you have throws Exception, you can't then catch that exception in that same method, which is what you are doing in that method below. 
From: https://rollbar.com/guides/java-throwing-exceptions/
"Throws is a keyword used to indicate that this method could throw this type of exception. The caller has to handle the exception using a try-catch block or propagate the exception."
Since you as the caller are handling the error by catching the error, you can just remove the throws IOException for the getGroupFromId/getUserFromId method

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
nicorod-google: Okie dokie!

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private Entity getGroupFromId(
+    HttpServletResponse response, long groupId, DatastoreService datastore) throws IOException {
+    try {
+      return datastore.get(KeyFactory.createKey("Group", groupId));
+    } catch (EntityNotFoundException e) {
+      ErrorHandler.sendError(response, "Group does not exist.");
+      return null;
+    }
+  }
```
g-asweet, line 73: Looping back to what Nico said, this claims to throw an error, but you're catching it, so it cannot throw an error. Remove throws from here.

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private Entity getUserFromId(
+    HttpServletResponse response, String userId, DatastoreService datastore) throws IOException {
+    try {
+      return datastore.get(KeyFactory.createKey("User", userId));
+    } catch (EntityNotFoundException e) {
+      ErrorHandler.sendError(response, "User does not exist.");
+      return null;
+    }
+  }
```
g-asweet, line 83: Remove `throws IOException`, since you're catching it.

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
g-asweet: ^ although if https://github.com/googleinterns/step100-2020/blob/master/capstone/src/main/java/error/ErrorHandler.java

ErrorHandler.sendError DOES throw an exception (I guess in the final write line?), then yeah I suppose you can't remove the throws IOException

g-asweet (approved):

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
nicorod-google: Right but since she is catching the exception in those methods, wouldn't she not need the "throws IOException" as part of those methods? Though definitely for this doPost method I think it's necessary

nicorod-google (approved):

PR #19, opened 2020-07-09
Servlet and Objects Testing

lxcyqx:
Wrote Junit tests, including following test classes:
- ChallengeServletTest
- MarkCheckboxTest
- PollServletTest
- ChallengeTest
- OptionTest
- PollTest

Also added OptionsAndUserVotedOptions class to refactor PollServlet class.

These tests are based off of code from challenge-checkbox branch but I'm putting base as master instead of challenge-checkbox because having base as challenge-checkbox still results in other files changed


capstone/src/main/java/com/google/sps/Objects/Challenge.java
```diff
@@ -1,57 +1,128 @@
...
+ *
+ * @author lucyqu
+ */
 public final class Challenge {
 
   private final String challengeName;
   private final long dueDate;
   private final Badge badge;
-  private final ArrayList<String> usersCompleted;
+  private final List<String> usersCompleted;
```
nicorod-google, line 18: I believe this is part of the previous PR, but I'm curious as to the reason why you decided to switch from an ArrayList to a List?

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Test
+  public void getDueDateTest() {
+    assert challenge.getDueDate() == 123456;
+  }
+
+  @Test
+  public void getChallengeName() {
+    assertEquals(challenge.getChallengeName(), "Run 3 miles");
```
himaniyadav: nit: I think it might be good to declare these (and any other challenge fields) as constants up at the top, so the assert would look like `challenge.getDueDate() == DUE_DATE`. just personal preference but I think it makes the tests a bit clearer. and if for some reason you want a different challenge name, due date, id, etc, you only have to change everything in one place

capstone/src/test/java/com/google/sps/servlets/ChallengeServletTest.java
```diff
@@ -0,0 +1,168 @@
...
+            NEW_CHALLENGE, /* challenge name*/
+            dueDate, /* due date */
+            null, /* badge */
+            new ArrayList<String>(), /* users completed */
+            id /* id of challenge */);
+    Challenge returnedChallenge = Challenge.fromEntity(entity);
+
+    assertEquals(challenge.getChallengeName(), returnedChallenge.getChallengeName());
+    assert challenge.getUsersCompleted().size() == returnedChallenge.getUsersCompleted().size();
+    assert challenge.getDueDate() == returnedChallenge.getDueDate();
```
himaniyadav: You could override the challenge .equals() method to compare these instead of manually comparing each field. Or you could convert both challenge objects to JSON strings and compare the JSON.

capstone/src/test/java/com/google/sps/servlets/ChallengeServletTest.java
```diff
@@ -0,0 +1,168 @@
...
+    assertTrue(response.contains("Oops an error happened!"));
+  }
+
+  @Test
+  public void doPost_userLoggedIn() throws IOException, EntityNotFoundException {
+    when(mockRequest.getParameter("name")).thenReturn(NEW_CHALLENGE);
+
+    challengeServlet.doPost(mockRequest, mockResponse);
+    // id of Challenge increments for each new Challenge that is added
+    Key challengeKey = KeyFactory.createKey("Challenge", CHALLENGE_ID);
```
himaniyadav, line 136: Hmm what do you mean by this? Is this a feature that you wrote in the ChallengeServlet class or a feature of the LocalTestService ?

capstone/src/test/java/com/google/sps/servlets/PollServletTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+  @Test(expected = EntityNotFoundException.class)
+  public void doPost_userNotLoggedIn() throws IOException, EntityNotFoundException {
+    helper.setEnvIsLoggedIn(false);
+    when(mockRequest.getParameter("text")).thenReturn(NEW_OPTION);
+
+    pollServlet.doPost(mockRequest, mockResponse);
+    // id of Option increments for each new Option that is added
+    Key optionKey = KeyFactory.createKey("Option", OPTION_TEXT.size() + 1);
+    Entity optionEntity = datastore.get(optionKey);
```
himaniyadav, line 157: add a comment here specifying that this line should trigger the error

capstone/src/test/java/com/google/sps/servlets/ChallengeServletTest.java
lxcyqx: It's a feature of LocalTestService so if there is one thing in LocalTestService datastore it will have id of 1 and the following one that's added will have id of 2

himaniyadav: Ah, got it

capstone/src/main/java/com/google/sps/servlets/AuthenticatedServlet.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+
+import error.ErrorHandler;
+
+public abstract class AuthenticatedServlet extends HttpServlet {
```
g-asweet: Looks great!

capstone/src/main/java/com/google/sps/servlets/MarkCheckboxServlet.java
```diff
@@ -0,0 +1,128 @@
...
+  @Override
+  public void doPost(String userId, HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
+    String idString = request.getParameter("id");
+    boolean isChecked = Boolean.parseBoolean(request.getParameter("checked"));
+    // type representing whether checkbox is for Option or for Challenge
+    String type = request.getParameter("type");
+    long id = this.parseToLong(response, idString);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    //    String userId = this.getUserId(response);
```
g-asweet: Can this line be removed?

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Test
+  public void getIsCompletedTest() {
+    assertFalse(challenge.getHasUserCompleted("1"));
+    challenge.addCompletedUser("1");
+    assertTrue(challenge.getHasUserCompleted("1"));
+    challenge.addCompletedUser("2");
+    challenge.addCompletedUser("3");
+    assertFalse(challenge.getHasUserCompleted("4"));
+  }
```
g-asweet: I think this should be two tests

`getIsCompletedTest_hasUser()`
`getIsCompletedTest_doesNotHaveUser()`

capstone/src/test/java/com/google/sps/objects/OptionTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    for (int i = 0; i < votesToAdd.size(); i++) {
+      option.addVote(votesToAdd.get(i));
+    }
+
+    assert option.getVotes().size() == 2;
+    List<String> votes = option.getVotes();
+    for (int i = 0; i < votes.size(); i++) {
+      assertEquals(option.getVotes().get(i), votesToAdd.get(i));
+    }
```
g-asweet: I would prefer to not see any for-loops if possible. This is great code for production code, test code often is quite the opposite when it comes to variables and reuse of values.

```java
option.addVote("1");
option.addVote("234");

assert option.getVotes().size() == 2;
assertEquals(option.getVotes().get(0), "1");
assertEquals(option.getVotes().get(0), "234");
```

Using the repeated magic numbers in-line is much easier to follow exactly what is being tested and does not obfuscate a failing test case behind a variable/a list/a for-loop

capstone/src/test/java/com/google/sps/objects/OptionTest.java
```diff
@@ -0,0 +1,107 @@
...
+    datastore.put(entity);
+
+    Option returnedOption = Option.fromEntity(entity);
+    List<String> returnedVotes = (ArrayList<String>) entity.getProperty("votes");
+
+    assertEquals(returnedOption.getText(), OPTION_NAME);
+    assert returnedVotes.size() == 3;
+    for (int i = 0; i < returnedVotes.size(); i++) {
+      assertEquals(returnedVotes.get(i), votes.get(i));
+    }
```
g-asweet: Please remove this for-loop, in favor or in-line magic numbers.

capstone/src/test/java/com/google/sps/servlets/MarkCheckboxTest.java
```diff
@@ -0,0 +1,135 @@
...
+    assertEquals(challenge.getUsersCompleted().get(0), USER_ID);
+  }
+
+  @Test(expected = EntityNotFoundException.class)
+  public void doPost_userNotLoggedIn() throws IOException, EntityNotFoundException {
+    helper.setEnvIsLoggedIn(false);
+    when(mockRequest.getParameter("name")).thenReturn(NEW_CHALLENGE);
+
+    challengeServlet.doPost(mockRequest, mockResponse);
+
```
g-asweet: Tests are usually preferred to be in three blocks:

// Arrange

// Act

// Assert

Right now, it's hard for me to understand why you have to `challengeServlet.doPost` before doing your `markCheckboxServlet.doPost`. I recommend one of two suggestions:

1. Seeing multiple blocks here, as well as a couple of repeated lines of code in multiple tests suggests that adding a helper function for testing may make this read more clearly:

```java
helper.setEnvIsLoggedIn(false);
setupChallenge();
when(mockRequest.getParameter("id")....
```

2. If you have to call another file's doPost in order to make your unit test work, it may not be a unit test. Consider mocking some behavior that MarkCheckbox is expecting from the database.

g-asweet (changes requested):

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+    challenge = null;
+  }
+
+  @Test
+  public void getDueDateTest() {
+    assert challenge.getDueDate() == DUE_DATE;
```
nicorod-google: I suggest changing this to assertEquals to keep consistency with the other tests

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  @Test
+  public void getBadgeTest() {
+    assertEquals(challenge.getBadge(), badge);
+  }
+
+  @Test
+  public void getUsersCompletedTest() {
+    assert challenge.getUsersCompleted().size() == 0;
```
nicorod-google: Same comment about assertEquals

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  @Test
+  public void addCompletedUserTest() {
+    challenge.addCompletedUser("1");
+    challenge.addCompletedUser("2");
+
+    List<String> completedUsers = challenge.getUsersCompleted();
+
+    assert completedUsers.size() == 2;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
```diff
@@ -0,0 +1,134 @@
...
+    String challengeName = CHALLENGE_NAME;
+    entity.setProperty("name", challengeName);
+    entity.setProperty("votes", new ArrayList<String>());
+    entity.setProperty("dueDate", DUE_DATE);
+
+    Challenge returnedChallenge = Challenge.fromEntity(entity);
+
+    assertEquals(returnedChallenge.getChallengeName(), challengeName);
+    assert returnedChallenge.getUsersCompleted().size() == 0;
+    assert returnedChallenge.getDueDate() == DUE_DATE;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
```diff
@@ -0,0 +1,134 @@
...
+  public void toEntityTest() {
+    Entity entity = challenge.toEntity();
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    datastore.put(entity);
+    List<String> votes = (ArrayList<String>) entity.getProperty("votes");
+
+    assertEquals(entity.getProperty("name"), CHALLENGE_NAME);
+    assert votes.size() == 0;
+    assert (long) entity.getProperty("dueDate") == DUE_DATE;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
```diff
@@ -0,0 +1,134 @@
...
+            DUE_DATE, /* due date */
+            badge, /* badge */
+            new ArrayList<String>(), /* completed users */
+            ID /* challenge id */);
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+    challenge = null;
```
nicorod-google: I think this is unnecessary because the setUp is called before every test, therefore if a test changes challenge, it will be overridden to the state in setUp before the next test is run

capstone/src/test/java/com/google/sps/objects/OptionTest.java
```diff
@@ -0,0 +1,99 @@
...
+  public void getTextTest() {
+    assertEquals(option.getText(), OPTION_NAME);
+  }
+
+  @Test
+  public void addVoteGetVotesTest() {
+    option.addVote("1");
+    option.addVote("2");
+
+    assert option.getVotes().size() == 2;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/objects/OptionTest.java
```diff
@@ -0,0 +1,99 @@
...
+    assertEquals(option.getVotes().get(0), "1");
+    assertEquals(option.getVotes().get(1), "2");
+  }
+
+  @Test
+  public void getIdTest() {
+    assert option.getId() == 100;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/objects/OptionTest.java
```diff
@@ -0,0 +1,99 @@
...
+  @Test
+  public void toEntityTest() {
+    Entity entity = option.toEntity();
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(entity);
+    ArrayList<String> votes = (ArrayList<String>) entity.getProperty("votes");
+
+    assertEquals(entity.getProperty("text"), OPTION_NAME);
+    assert votes.size() == 0;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/objects/OptionTest.java
```diff
@@ -0,0 +1,99 @@
...
+    votes.add("3");
+    entity.setProperty("votes", votes);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(entity);
+
+    Option returnedOption = Option.fromEntity(entity);
+    List<String> returnedVotes = (ArrayList<String>) entity.getProperty("votes");
+
+    assertEquals(returnedOption.getText(), OPTION_NAME);
+    assert returnedVotes.size() == 3;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/objects/PollTest.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    poll = null;
+  }
+
+  @Test
+  public void getOptionTest() {
+    assert poll.getOptions().size() == 0;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/objects/PollTest.java
```diff
@@ -0,0 +1,54 @@
...
+  @Test
+  public void addOptionTest() {
+    Option option = new Option(ID, OPTION_NAME, new ArrayList<String>());
+    Option option2 = new Option(ID, OPTION_NAME, new ArrayList<String>());
+    poll.addOption(option);
+    poll.addOption(option2);
+
+    List<Option> options = poll.getOptions();
+
+    assert options.size() == 2;
```
nicorod-google: assertEquals

capstone/src/test/java/com/google/sps/servlets/ChallengeServletTest.java
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Sets due date to midnight, 7 days from when challenge is posted.
+   *
+   * @return due date in milliseconds
+   */
+  private long getDueDate() {
+    LocalDateTime now = LocalDateTime.now();
+    LocalDateTime dueDate = now.plusDays(7).withHour(23).withMinute(59).withSecond(59).withNano(0);
+    return Timestamp.valueOf(dueDate).getTime();
+  }
```
nicorod-google: You typically don't want to rely on other services to run correctly for your tests. Importantly, how can you guarantee the dueDate created for the entity will be the same as the one created in doPost_userLoggedIn? Instead of using LocalDateTime, you should create a static DUE_DATE variable like you did it in MarkCheckboxTest.

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/Challenge.java
lxcyqx: I learned in school that it's better to initialize as List rather than ArrayList because it's more generic and noticed that here I left it as ArrayList

capstone/src/test/java/com/google/sps/servlets/ChallengeServletTest.java
lxcyqx: I put this logic in here so that I could check that the due date inputted into the database is calculated correctly. Should I just not assert whether the due dates are equal then?

nicorod-google: I see. In that case, rather than needing to call the method twice (in createChallenge & the test) you should create the constant variable above and use that throughout.

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -18,47 +15,46 @@
...
 
   @Override
-  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+  public void doPost(String userId, HttpServletRequest request, HttpServletResponse response)
+      throws IOException {
     String challengeName = request.getParameter("name");
-    LocalDateTime dueDate = this.getDueDate(LocalDateTime.now());
-    long dueDateMillis = Timestamp.valueOf(dueDate).getTime();
+    Time time = new Time();
+    long dueDate = Challenge.getDueDate(time);
```
nicorod-google: nit: inline the Time instantiation instead of assigning it to a variable that you only use here

capstone/src/main/java/com/google/sps/servlets/PollServlet.java
```diff
@@ -16,29 +15,49 @@
...
     Query query = new Query("Option").addSort("timestamp", SortDirection.ASCENDING);
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     PreparedQuery results = datastore.prepare(query);
+    OptionsAndUserVotedOptions optionsAndVotedOptions =
+        this.getOptionsAndVotedOptions(results, userId);
+    List<Option> options = optionsAndVotedOptions.getOptions();
+    List<Long> votedOptions = optionsAndVotedOptions.getVotedOptions();
+    // Sort list of options based on number of votes
+    Collections.sort(options, new OptionsComparator());
+    PollResponse pollResponse = new PollResponse(options, votedOptions, userId);
```
nicorod-google: I want to apologize since I know we talked about this during the our daily meeting, however I think we had a lack of context when talking about this. The OptionsAndVotedOptions here is kind of redundant since they are just used to pass into & create a PollResponse. So instead of needing to create the OptionsAndUserVotedOptions object, your helper method should create & return a PollResponse. 
Instead of: 
`
private OptionsAndUserVotedOptions getOptionsAndVotedOptions(PreparedQuery results, String userId) {
     ...
     return new OptionsAndUserVotedOptions(options, votedOptions);
}
`
you can have 
`
private PollResponse buildPollResponse(PreparedQuery results, String userId) {
     ...
     return new PollResponse(options, votedOptions, userId);
}
`
I know you also need to sort the options list so you can just add that in the helper method after building the list.
So that would leave you with:
`
PollResponse pollResponse = buildPollResponse(results, userId);
`

capstone/src/test/java/com/google/sps/objects/TimeTest.java
```diff
@@ -0,0 +1,34 @@
...
+  @Before
+  public void setUp() throws IOException {
+    MockitoAnnotations.initMocks(this);
+  }
+
+  @Test
+  public void getDueDateTest() {
+    LocalDateTime now = LocalDateTime.of(2020, Month.JULY, 10, 10, 30, 40);
+    when(mockedTime.getNow()).thenReturn(now);
+    long actualDueDate = 1595044799000L;
```
nicorod-google: nit: Make this a file constant

nicorod-google (changes requested):

capstone/src/test/java/com/google/sps/objects/ChallengeTest.java
g-asweet: Agreed, no need to null it out on tearDown. I almost never use a teardown unless you really need to destroy/deconstruct something, since the JVM will automatically garbage-collect the original object once your replace the challenge

g-asweet (approved):

nicorod-google (approved):

PR #21, opened 2020-07-09
Edit profile functionality

himaniyadav:
- Added the ability for a user to edit their profile information
- Added "Join Group" and "Create Group" buttons on profile page (not functional) in preparation for next week

Screenshots of profile before edit, edit modal, and profile after edit:
image
image
image

### DIFF
--- a/capstone/src/main/java/com/google/sps/servlets/EditProfileServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/EditProfileServlet.java
@@ -33,13 +33,7 @@ public class EditProfileServlet extends HttpServlet {
     String phone = request.getParameter("phone");
     ArrayList<String> interests = getInterests(request);
 
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    
-    Entity userEntity = getExistingUser(response, datastore);
-    
-    User user = getUpdatedUser(userEntity, first, last, email, phone, interests);
-
-    datastore.put(user.toEntity());
+    updateProfile(first, last, email, phone, interests, response);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


capstone/src/main/java/com/google/sps/servlets/EditProfileServlet.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+    
+    User user = User.fromEntity(userEntity);
+    user.setFirstName(first);
+    user.setLastName(last);
+    user.setEmail(email);
+    user.setPhoneNumber(phone);
+    user.setInterests(interests);
+
+    datastore.put(user.toEntity());
```
g-asweet: This is a broad comment I made in a 1:1 with Lucy, but I'll try to reiterate for everyone: @anikabagga @lxcyqx 

Due to how you have to structure your servlets to match a very specific interface, the only way I can figure out what each servlet does is:
1. Read the name of the servlet, to guess what it's supposed to do on a Post/Get
2. Look all the lines in 20-40 line function called doPost, and read over all the parts of it to figure out what data it's getting in the request and what the outcome of that request is.

I think to minimize the cognitive load around figuring out what each servlet does, I would like to request any action that's being taken by the server exist within a well-named function. This doesn't mean just move all this logic into a helper function and call it, leaving doPost to be a one line function, but I would say that doPost can get a request, format data however it needs to be formatted, do any logic tests/checking/validation that you have a valid post, and once you've done those things, consider calling a well-named helper function.

In this case, a `doPost` on `editProfileServlet` should likely call a helper function called `updateProfile`, so that I can easily read within just those 3 names, what servlet it is, is it a get or a post, and ultimately what action the server is making in response.

g-asweet (approved):

PR #22, opened 2020-07-10
Object Testing 

anikabagga:
Junit tests for object files:

CommentTest.java
PostTest.java
MemberResponseTest.java


capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -49,6 +49,21 @@ public static Post getPostEntity(Entity entity) {
...
+    Post post = (Post) other;
+    return timestamp == post.timestamp &&
+      authorId.equals(post.authorId) &&
+      postText.equals(post.postText) &&
+      challengeName.equals(post.challengeName) &&
+      img.equals(post.img) &&
+      likes.containsAll(post.likes) &&
+      comments.containsAll(post.comments);
+  }
+
```
lxcyqx, line 89: I think Andrew suggested that instead of doing this you can compare the json for testing purposes. This is so that if the class is updated, you don't have to then also make sure this equals method is updated

capstone/src/test/java/com/google/sps/servlets/CommentTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Test
+  public void getCommentEntityTest() {
+    EmbeddedEntity entity = new EmbeddedEntity();
+    entity.setProperty("userId", USER_ID);
+    entity.setProperty("commentText", COMMENT_TEXT);
+    entity.setProperty("timestamp", TIMESTAMP);
+    Comment returnedComment = Comment.getCommentEntity(entity);
+
+    String jsonRetrieved = new Gson().toJson(returnedComment);
```
lxcyqx: Since you're comparing the json here is there a reason to have that equals method in Comment.java?

capstone/src/main/java/com/google/sps/Objects/Comment.java
```diff
@@ -22,6 +22,24 @@ public static Comment getCommentEntity(EmbeddedEntity entity) {
...
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (other == null) return false;
+    if (other == this) return true;
+    if (!(other instanceof Comment)) return false;
+    Comment comment = (Comment) other;
+    return commentText.equals(comment.commentText) &&
+      userId.equals(comment.userId);
```
g-asweet: @lxcyqx @himaniyadav

I would challenge this equality check, and others. If I comment twice on two different challenges or even on the same challenge with the same text, this equality check is claiming they're the same. I don't know exactly how equals plans to be used throughout the entire server, but I could see equals doing one of two things:

1) Just check if the object ID's are the same. The benefit of this is that if I post something and retrieve it later from the database after somebody has liked it, we can still call these posts as "equal" even though not all fields are up to date. Downside is you guys will have to add some additional logic around to/from entity to better manage ID's.
2) Check the Object ID is the same or that the timestamp is the same, as well as all other fields. The downside is what you're already seeing, that you have to check all fields manually and update it as you go. The use case here is that you'd now have 2 kinds of equals: obj1.id === obj2.id and obj1.equals(obj2) which would be used for different cases.

EITHER WAY I think that a lot of what I'm seeing in the equality checks may be completely invalidated without using something like an ID or timestamp to actually check for uniqueness.

capstone/src/main/java/com/google/sps/Objects/Post.java
g-asweet: Yeah I think using json string comparison is a valid way to check comparison without having to update this function every time you add a field. You may have to make a temporary object and modify it to make sure you're only comparing fields which are REQUIRED for equality (i.e. I think you had cases where a property could be missing, could be null, or could be an empty array and then you'd have to determine which of those cases SHOULD be considered equal).

capstone/src/main/java/com/google/sps/Objects/response/MemberResponse.java
```diff
@@ -58,6 +58,30 @@ public static MemberResponse fromEntity(Entity entity, boolean includeBadges) {
...
+      return profilePic.equals(memResponse.profilePic) &&
+        firstName.equals(memResponse.firstName) &&
+        lastName.equals(memResponse.lastName) &&
+        userId.equals(memResponse.userId) &&
+        badges.containsAll(memResponse.badges);
+    }
+    return profilePic.equals(memResponse.profilePic) &&
+      firstName.equals(memResponse.firstName) &&
+      lastName.equals(memResponse.lastName) &&
+      userId.equals(memResponse.userId);
```
g-asweet: If you end up going the JSON route, you may want to create temporary objects where the badges field would ALWAYS be an array (instead of null) then convert them to json, then string compare

capstone/src/test/java/com/google/sps/servlets/CommentTest.java
g-asweet: I would only point out that the tests should not define your production API's. But if nobody is actually USING the `equals` functions, then you all certainly don't have to define them.

capstone/src/test/java/com/google/sps/servlets/MemberResponseTest.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @Test
+  public void fromEntityTest() {
+    Entity entity = new Entity("MemberResponse");
+    entity.setProperty("profilePic", PROFILE_PIC);
+    entity.setProperty("firstName", FIRST_NAME);
+    entity.setProperty("lastName", LAST_NAME);
+    entity.setProperty("userId", USER_ID);
+    MemberResponse returnedMemberResponse = MemberResponse.fromEntity(entity, false /*includeBadges*/);
```
g-asweet: When I see something like `/* includeBadges */ false`, it makes me think there should be a test-case for `/* includeBadges */ true`

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+    assert testLikes.containsAll(post.getLikes());
+  }
+
+  @Test
+  public void addCommentTest() {
+    Comment testComment1 =  
+      new Comment(
+        4324344, /* userId */ 
+        "a great comment", /* commentText */ 
+        "123123123" /* userId */ );
```
g-asweet:
```java
  /* userId */ 4324344,
  /* commentText */ "a great comment",
  /* userId */ "123123123"
```

reads a bit cleaner to me, because then you can read all the parameter names in a column.

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+  public void getPostEntityTest() {
+    Entity entity = new Entity("Post");
+    entity.setProperty("authorId", AUTHOR_ID);
+    entity.setProperty("timestamp", TIMESTAMP);
+    entity.setProperty("postText", POST_TEXT);
+    entity.setProperty("challengeName", CHALLENGE_NAME);
+    entity.setProperty("img", IMG);
+    entity.setProperty("likes", new ArrayList<String>());
+    entity.setProperty("comments", new ArrayList<Comment>());
+    Post returnedPost = Post.getPostEntity(entity);
```
g-asweet: Can you put a new line above `getPostEntity`? Gives this test a clean 3 block structure

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+    likes = post.getLikes();
+
+    assert likes.size() == 1;
+    assert likes.contains("test user 2");
+
+    post.removeLike("test user 2");
+    likes = post.getLikes();
+
+    assert likes.size() == 0;
+  }
```
g-asweet: This should be 2 3 block tests:

`testRemoveLike` and `testRemoveLikeMultiple`

where you call removeLike once, or twice, and then check the value.

Having multiple places where you're asserting is less clean and obvious what's going wrong when you debug a failing test.

g-asweet (changes requested):

capstone/src/test/java/com/google/sps/servlets/MemberResponseTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Test
+  public void getStringURLTest() {
+    assertEquals(memberResponse.getStringURL(), PROFILE_PIC);
+  }
+
+  @Test
+  public void getBadgesTest() {
+    assert memberResponse.getBadges().size() == 0;
+    badge = new Badge(/* challenge name */ "workout", /* icon */ null, /* timestamp */ 11111111);
```
nicorod-google: The timestamp argument appears to be the same as the USER_ID constant above. Can you make the arguments here constants (except for the null icon) and perhaps change the timestamp or user id to be different values?

capstone/src/test/java/com/google/sps/servlets/MemberResponseTest.java
```diff
@@ -0,0 +1,97 @@
...
+    memberResponse.getBadges().add(badge);
+    assert memberResponse.getBadges().size() == 1;
+  }
```
nicorod-google: Rather than asserting multiple states in a single test, you should split them out into different tests i.e. something like getBadgesTest_noBadges and getBadgesTest_hasBadges

capstone/src/test/java/com/google/sps/servlets/MemberResponseTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Test
+  public void getStringURLTest() {
+    assertEquals(memberResponse.getStringURL(), PROFILE_PIC);
+  }
+
+  @Test
+  public void getBadgesTest() {
+    assert memberResponse.getBadges().size() == 0;
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/MemberResponseTest.java
```diff
@@ -0,0 +1,97 @@
...
+    badge = new Badge(/* challenge name */ "workout", /* icon */ null, /* timestamp */ 11111111);
+    memberResponse.getBadges().add(badge);
+    assert memberResponse.getBadges().size() == 1;
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @Test
+  public void getChallengeNameTest() {
+    assertEquals(post.getChallengeName(), CHALLENGE_NAME);
+  }
+
+  @Test 
+  public void getCommentsTest() {
+    assert post.getComments().size() == 0;
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  @Test
+  public void getLikesTest() {
+    assert post.getLikes().size() == 0;
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+    post.addLike("user 1");
+    post.addLike("user 2");
+    HashSet<String> testLikes = new HashSet<String>();
+    testLikes.add("user 1");
+    testLikes.add("user 2");
+    
+    assert testLikes.containsAll(post.getLikes());
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+        "123123123" /* userId */ );
+    Comment testComment2 = 
+      new Comment(
+        55555555, /* userId */ 
+        "another great comment", /* commentText */ 
+        "09090909" /* userId */ );
+    post.addComment(testComment1);
+    post.addComment(testComment2);
+
+    assert post.getComments().size() == 2;
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @Test
+  public void addLikeTest() {
+    post.addLike("user 1");
+    post.addLike("user 2");
+    HashSet<String> likes = post.getLikes();
+
+    assert likes.size() == 2;
+    assert likes.contains("user 1");
+    assert likes.contains("user 2");
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  @Test
+  public void removeLikeTest() {
+    post.addLike("test user 1");
+    post.addLike("test user 2");
+    HashSet<String> likes = post.getLikes();
+    assert likes.size() == 2;
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    post.removeLike("test user 1");
+    likes = post.getLikes();
+
+    assert likes.size() == 1;
+    assert likes.contains("test user 2");
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

capstone/src/test/java/com/google/sps/servlets/PostTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    post.removeLike("test user 2");
+    likes = post.getLikes();
+
+    assert likes.size() == 0;
```
nicorod-google: nit: Change this to assertEquals to make it consistent with the other asserts

nicorod-google (changes requested):

capstone/src/test/java/com/google/sps/servlets/MemberResponseTest.java
anikabagga: sounds good, I'll add the getBadgesTest_hasBadges once I implement badges

capstone/src/test/java/com/google/sps/servlets/MemberResponseTest.java
anikabagga: Will include two separate tests once I implement badges for members.

capstone/src/test/java/com/google/sps/servlets/CommentTest.java
anikabagga: I'm using the comment equals function in PostTest, but can also change it to a json comparison to keep things consistent.

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -69,66 +68,24 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     String challengeName = "Challenge Name";
     String img = getUploadedFileUrl(request, "image");
-    ArrayList<String> likes = new ArrayList<>();
+    HashSet<String> likes = new HashSet<>();
     ArrayList<Comment> comments = new ArrayList<>();
 
     // Creates entity with submitted data and add to database
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(createPostEntity(timestamp, authorName, postText, challengeName, img, likes, comments));
+    Post post = new Post(0, authorName, postText, comments, challengeName, System.currentTimeMillis(), img, likes);
```
nicorod-google, line 80: This looks like every post has an ID of 0, is that intended?

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -36,18 +36,17 @@
...
+
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
-    String userId = "";
-    UserService userService = UserServiceFactory.getUserService();
-    if (userService.isUserLoggedIn()) {
-      userId = userService.getCurrentUser().getUserId();
-    }
+    String userId = this.getUserId(response);
```
nicorod-google: Same comment from other file: Lucy made an AuthenticatedServlet that you can extend instead of the HttpServlet that will do the user Authentication for you: https://github.com/googleinterns/step100-2020/blob/master/capstone/src/main/java/com/google/sps/servlets/AuthenticatedServlet.java
So you just need to make sure to extend the AuthenticatedServlet and you can remove an user login related code. Also note the doPost/doGet change to give you access to the userID

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
```diff
@@ -31,6 +35,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     UserService userService = UserServiceFactory.getUserService();
     if (userService.isUserLoggedIn()) {
       userId = userService.getCurrentUser().getUserId();
+    } else {
+      errorHandler.sendError(response, "User is not logged in.");
     }
```
nicorod-google: Lucy made an AuthenticatedServlet that you can extend instead of the HttpServlet that will do the user Authentication for you: https://github.com/googleinterns/step100-2020/blob/master/capstone/src/main/java/com/google/sps/servlets/AuthenticatedServlet.java
So you just need to make sure to extend the AuthenticatedServlet and you can remove an user login related code. Also note the doPost/doGet change to give you access to the userID

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: I'm using the 0 as a placeholder id. Right after I'm creating a post entity, in which the id will be replaced with an id generated by the database.

capstone/src/main/java/com/google/sps/servlets/CommentsServlet.java
anikabagga: Sounds good, added to my servlets

capstone/src/main/java/com/google/sps/Objects/Post.java
```diff
@@ -46,9 +47,46 @@ public static Post getPostEntity(Entity entity) {
...
   }
 
+  private static void createCommentObjectList(ArrayList<Comment> comments, Entity entity) {
+    ArrayList<EmbeddedEntity> commentEntitys = (ArrayList<EmbeddedEntity>) entity.getProperty("comments");
+    for (EmbeddedEntity comment: commentEntitys) {
+      comments.add(Comment.getCommentEntity(comment));
+    }
+  }
+
+  public Entity createPostEntity() {
```
himaniyadav: I can't comment on line 39 `public static Post getPostEntity(Entity entity)` because you didn't change it, but I think it makes sense to rename that method to something like getFromPostEntity (or just fromEntity) for consistency/clarity

capstone/src/main/java/com/google/sps/Objects/response/MemberResponse.java
anikabagga: If I retrieve an entity from the database with an empty list of badges, the property badges of the entity would return null. Added this check in case I want to compare two objects after getting the entity from the database.

capstone/src/main/java/com/google/sps/Objects/Comment.java
anikabagga: I'm using the equality checks more as a way to make sure the correct data is being posted/retrieved, but not necessarily checking for uniqueness. Since the timestamp and object ID's are being generated when an entity is created/put in the database, I haven't found a way to mandate that. Could I add a test to make sure two objects with the exact same data have two different ID's for uniqueness?

nicorod-google (approved):

capstone/src/main/java/com/google/sps/Objects/Comment.java
g-asweet: If you're checking for data retrieval correctness, my argument would be that this logic should exist in the unit tests instead of a production-available function then. Obviously if you're using `.equals()` somewhere in production you can keep the logic here, but your comment above seems more like unit-test logic than a method that should be usable in production.

capstone/src/main/java/com/google/sps/Objects/response/MemberResponse.java
g-asweet: Could you refactor this so you're only checking profilePic, firstName, lastName, userId in one place in this function so that it's easier to maintain when you add/remove fields?

g-asweet (dismissed):

capstone/src/main/java/com/google/sps/Objects/response/MemberResponse.java
anikabagga: sure! changed

nicorod-google (approved):

PR #23, opened 2020-07-10
User profile servlet tests

himaniyadav:
Added JUnit tests for the following servlets:
- UserDataServlet
- UserGroupsServlet
- EditProfileServlet


capstone/src/main/java/com/google/sps/Objects/response/UserGroupResponse.java
```diff
@@ -34,7 +34,9 @@ public static UserGroupResponse fromEntity(Entity entity) {
     String groupName = (String) entity.getProperty("groupName");
     String headerImg = (String) entity.getProperty("headerImg");
     // TODO: fix below code, won't work with Challenges as EmbeddedEntities of Groups
-    ArrayList<Challenge> challenges = (ArrayList<Challenge>) entity.getProperty("challenges");
+    ArrayList<Challenge> challenges = (entity.getProperty("challenges") == null)
+        ? new ArrayList<>()
+        : (ArrayList<Challenge>) entity.getProperty("challenges");
```
g-asweet, line 39: I like this, it's cleaner than null vs array

capstone/src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,138 @@
...
+  @Test
+  public void doGet_retrieveUserData() throws Exception {
+    userDataServlet.doGet(mockRequest, mockResponse);
+    String response = responseWriter.toString();
+
+    String expectedResponse = new Gson().toJson(USER_1);
+
+    // remove any whitespace from JSON
+    response = response.replaceAll("\\s", "");
+    expectedResponse = expectedResponse.replaceAll("\\s", "");
```
g-asweet: I'm curious, are the json object generating different whitespace? It may be helpful to remove this whitespace stripping logic into a helper function, maybe something like `assertEqualJson`. I really don't like Util files, but I do see this is a problem in multiple test files, so it could be something like a `TestUtils` file.

capstone/src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,138 @@
...
+    
+    assertEquals(response, expectedResponse);
+  }
+
+  @Test
+  public void doGet_userNotLoggedIn() throws Exception {
+    helper.setEnvIsLoggedIn(false);
+    userDataServlet.doGet(mockRequest, mockResponse);
+    String response = responseWriter.toString();
+    assertThat(response).contains("error");
```
g-asweet: This test (and others) should attempt to follow a 3 block structure.

https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80

g-asweet (dismissed):

capstone/src/test/java/com/google/sps/servlets/UserDataServletTest.java
himaniyadav: Yeah, not sure why but I think in the response there's sometimes an extra newline or carriage return that causes the equality tests to fail. I'll put it in a util file

nicorod-google (approved):
I agree with Andrew's comments, but otherwise everything lgtm

PR #24, opened 2020-07-10
Display badges on user profile

himaniyadav:
Added the proper functionality to display badges on a user's profile page.
- Badge fromEntity() method
- User getBadgeList() method
- frontend

Screenshot of UI (using hardcoded badges):
image


capstone/src/main/java/com/google/sps/Objects/User.java
```diff
@@ -171,4 +175,23 @@ public Entity toEntity() {
...
+  private static LinkedHashSet<Badge> getBadgeList(LinkedHashSet<Long> badgeIds) { 
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    LinkedHashSet<Badge> badges = new LinkedHashSet<>();
+    for (long badgeId : badgeIds) {
+      Key badgeKey = KeyFactory.createKey("Badge", badgeId);
+      Badge badge = null;
+      try { 
+        badge = Badge.fromEntity(datastore.get(badgeKey));
+      } catch(EntityNotFoundException e) {
+        // what to do here?
```
lxcyqx: are we doing the alert thing on the client side? If so we would need a field in the json that contains the exception messages

g-asweet: Can you break this up a little bit to only wrap what can throw an error? I think only the `datastore.get()` is problematic, right?

g-asweet: Also a broader comment: I'm guessing you guys will primarily expect 2 different kinds of errors:
1. Somebody maliciously attacking your API. You should have failsafes against invalid values etc.
2. Somebody left the page open too long, and the clientside data is now invalid because the database changed. (i.e. trying to comment on a post which has been deleted)

Since you should definitely ensure that you're protecting yourself against the first case, but it's not the primary use case that users can expect, I would consider adding a warning that explains to the user what happened ("That post no longer exists") with a reload page button on error or something, since it's very likely reloading the page will solve a users' problem, right?

g-asweet (dismissed):

capstone/src/main/java/com/google/sps/Objects/User.java
himaniyadav: Yup, that makes sense.

capstone/src/main/java/com/google/sps/Objects/Badge.java
```diff
@@ -1,28 +1,42 @@
...
+
+  /** Given a Badge entity, returns a Badge object. */
+  public static Badge fromEntity(Entity entity) {
+    long badgeId = (long) entity.getProperty("badgeId");
+    String challengeName = (String) entity.getProperty("challengeName");
+    String icon = (String) entity.getProperty("icon");
+    long timestamp = (long) entity.getProperty("timestamp");
+    
+    Badge badge = new Badge(badgeId, challengeName, icon, timestamp);
+    return badge;
```
nicorod-google: nit: Make this one line

capstone/src/main/webapp/scripts/profile.js
```diff
@@ -111,7 +110,15 @@ function displayChallenges(groups) {
 
 /** Display the user's earned badges  */
 function displayBadges(badges) {
-
+  const badgeContainer = document.getElementById('badge-grid');
+  for (badge of badges) {
+    let badgeElement = Document.createElement('div');
+    badgeElement.setAttribute('id', badge.badgeId);
+    badgeElement.setAttribute('class', 'badge');
+    badgeElement.setAttribute('title', badge.challengeName);
```
nicorod-google: nit: Make a helper function for creating a badge

capstone/src/main/webapp/scripts/profile.js
```diff
@@ -111,7 +110,15 @@ function displayChallenges(groups) {
 
 /** Display the user's earned badges  */
 function displayBadges(badges) {
-
+  const badgeContainer = document.getElementById('badge-grid');
+  for (badge of badges) {
+    let badgeElement = Document.createElement('div');
```
nicorod-google: nit: lowercase "document"

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/User.java
himaniyadav: The issue in this specific case is that since this is an object file, I'm not sure how to send the response.

nicorod-google: Instead of catching the error here could you instead pass it along to the servlets by throwing the error and catching the error when you call fromEntity?

capstone/src/main/java/com/google/sps/servlets/EditProfileServlet.java
```diff
@@ -88,8 +88,14 @@ private Entity getExistingUser(HttpServletResponse response, DatastoreService da
...
-      ArrayList<String> interests) {
-    User user = User.fromEntity(entity);
+      ArrayList<String> interests, HttpServletResponse response) throws IOException {
+    User user;
+    try {
+      user = User.fromEntity(entity);
+    } catch (EntityNotFoundException e) {
+      ErrorHandler.sendError(response, "User entity error.");
+      return null;
+    }
```
nicorod-google: This is good, however I'm not a fan of the generic "User entity error". This is my fault since I told you to get rid of thr previous try/catch block, but could you add that back but instead throw a new EntityNotFoundException with a relevant message in getBadgeList and use that message from e here instead?

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/EditProfileServlet.java
himaniyadav: Okay! Just to clarify, instead of putting "throws EntityNotFoundException" in getBadgeList() header, I should directly throw the error with a message in a catch block in getBadgeList()?

nicorod-google (approved):

PR #27, opened 2020-07-14
Servlet Testing 

anikabagga:
- AllGroupMembersServletTest.java
- UpdateLikesServletTest.java
- CommentsServletTest.java
- GroupPostDataServletTest.java 
- GroupMemberServletTest.java (in progress)


capstone/src/main/java/com/google/sps/servlets/UpdateLikesServlet.java
```diff
@@ -31,6 +31,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     String userId = this.getUserId(response);
     Entity post = this.getPostFromId(response, postId, datastore);
+    if (post == null || userId.equals("")) return;
```
himaniyadav, line 33: In what case would userId = ""? 
I think if you make this extend AuthenticatedServlet you can guarantee that userId will have a valid value

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -36,27 +36,26 @@
...
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
-    String userId = "";
-    UserService userService = UserServiceFactory.getUserService();
-    if (userService.isUserLoggedIn()) {
-      userId = userService.getCurrentUser().getUserId();
-    }
-
+    String userId = this.getUserId(response);
+    if (userId.equals("")) return;
```
nicorod-google: In case this is not part of the previous PR, you can also use the AuthenticatedServlet here to get the userID instead

capstone/src/test/java/com/google/sps/servlets/AllGroupMembersServletTest.java
```diff
@@ -0,0 +1,117 @@
...
+    when(mockResponse.getWriter()).thenReturn(new PrintWriter(responseWriter));
+    allGroupMembersServlet = new AllGroupMembersServlet();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+    responseWriter = null;
+    datastore = null;
+    allGroupMembersServlet = null;
```
nicorod-google: These are unnecessary because setUp is called before every test so it will make sure to return everything to the state that you configured them in setUp

capstone/src/test/java/com/google/sps/servlets/CommentsServletTest.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  private void populateDatabase(DatastoreService datastore) {
+    // Add test data.
+    datastore.put(POST_1.createPostEntity());
+  }
+
+  @Test
+  public void doPost_addComment() throws Exception {
+    populateDatabase(datastore);
```
nicorod-google: You should call this once in the setUp instead of at the beginning of every test

capstone/src/test/java/com/google/sps/servlets/GroupMemberServletTest.java
```diff
@@ -0,0 +1,135 @@
...
+    when(mockRequest.getParameter("groupId")).thenReturn(Long.toString(GROUP_1_ID));
+
+    groupMemberServlet.doPost(mockRequest, mockResponse);
+    String response = responseWriter.toString();
+    assertTrue(response.contains("Oops an error happened!"));
+  }
+
+  @Test
+  public void doPost_invalidPost() throws IOException, EntityNotFoundException {
+    populateDatabase(datastore);
```
nicorod-google: You should call this once in the setUp instead of at the beginning of every test

capstone/src/test/java/com/google/sps/servlets/GroupPostDataServletTest.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  private void populateDatabase(DatastoreService datastore) {
+    // Add test data.
+    datastore.put(POST_1.createPostEntity());
+  }
+
+  @Test
+  public void doGet_userLoggedIn() throws Exception {
+    populateDatabase(datastore);
```
nicorod-google: You should call this once in the setUp instead of at the beginning of every test

capstone/src/test/java/com/google/sps/servlets/UpdateLikesServletTest.java
```diff
@@ -0,0 +1,166 @@
...
+    when(mockRequest.getParameter("liked")).thenReturn("true");
+
+    updateLikesServlet.doPost(mockRequest, mockResponse);
+    String response = responseWriter.toString();
+    assertTrue(response.contains("Oops an error happened!"));
+  }
+
+  @Test
+  public void doPost_invalidPost() throws IOException, EntityNotFoundException {
+    populateDatabase(datastore);
```
nicorod-google: You should call this once in the setUp instead of at the beginning of every test

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/Objects/Comment.java
```diff
@@ -22,6 +24,34 @@ public static Comment getCommentEntity(EmbeddedEntity entity) {
...
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (other == null) return false;
+    if (other == this) return true;
+    if (!(other instanceof Comment)) return false;
+    Comment comment = (Comment) other;
+    return commentText.equals(comment.commentText) &&
+      userId.equals(comment.userId);
```
g-asweet: General comment I've left elsewhere, but if I leave a comment with the same text on two different posts, you're claiming this is the same comment. Alternatively, if I comment on a post, update the text, and you check the `.equals()` of the same comment before and after I edit the comment (a not currently supported feature, I'll admit), you're claiming that they are not equal.

My leaning would likely be to just use the database id as an equality check, or at least include the timestamp if you're going to go this route. I could be convinced that the same comment before and after an edit should not be equal, but I'm not really convinced that if I comment on two different posts, or the same point twice in a row, that the comments should be equal.

capstone/src/main/java/com/google/sps/Objects/response/MemberResponse.java
```diff
@@ -58,6 +58,30 @@ public static MemberResponse fromEntity(Entity entity, boolean includeBadges) {
...
+        firstName.equals(memResponse.firstName) &&
+        lastName.equals(memResponse.lastName) &&
+        userId.equals(memResponse.userId) &&
+        badges.containsAll(memResponse.badges);
+    }
+    return profilePic.equals(memResponse.profilePic) &&
+      firstName.equals(memResponse.firstName) &&
+      lastName.equals(memResponse.lastName) &&
+      userId.equals(memResponse.userId);
+  }
```
g-asweet: Can you refactor this so that the profile pic, firstName, lastName, and userId equality checks all happen only once so this is easier to maintain if a field gets added or removed?

capstone/src/main/java/com/google/sps/Objects/Comment.java
g-asweet: This may not matter at all in the long-term, and you guys are attempting to wrap up today so take this comment with a grain of salt to avoid having to redo all of the work you've done

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -143,4 +100,13 @@ private String getUploadedFileUrl(HttpServletRequest request, String formInputEl
     }
     return blobKey;
   }
+
+  private String getUserId(HttpServletResponse response) throws IOException {
+  UserService userService = UserServiceFactory.getUserService();
```
g-asweet: the indentation of this line is incorrect

capstone/src/test/java/com/google/sps/objects/CommentTest.java
```diff
@@ -0,0 +1,92 @@
...
+    assertEquals(comment.getUser(), USER_ID);
+  }
+
+  @Test
+  public void getCommentEntityTest() {
+    EmbeddedEntity entity = new EmbeddedEntity();
+    entity.setProperty("userId", USER_ID);
+    entity.setProperty("commentText", COMMENT_TEXT);
+    entity.setProperty("timestamp", TIMESTAMP);
+    Comment returnedComment = Comment.getCommentEntity(entity);
```
g-asweet: Can you put a new line before this line to better get a 3 block structure to the test? It'll make it more clear that you're testing the `getCommentEntity` function.

capstone/src/test/java/com/google/sps/objects/CommentTest.java
```diff
@@ -0,0 +1,92 @@
...
+    ArrayList<EmbeddedEntity> allComments = new ArrayList<>();
+    allComments.add(entity);
+    Entity commentEntity = new Entity("Comment");
+    commentEntity.setProperty("comments", allComments);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(commentEntity);
+
+    ArrayList<EmbeddedEntity> commentEntitys = (ArrayList<EmbeddedEntity>) commentEntity.getProperty("comments");
+    assertEquals(commentEntitys.get(0).getProperty("commentText"), COMMENT_TEXT);
+    assertEquals(commentEntitys.get(0).getProperty("userId"), USER_ID);
```
g-asweet: I may be wrong and fundamentally misunderstand this test, so please don't be offended, but here are my take-aways:

1. toEntityTest calls toEntity on the very first line before configuring much about the test. I assume the comment is being setup in the `setUp`, but the structure of a test should be:

// Arrange - Setup the test as you need, show what inputs are required and are being tested. Anything outside of this scope can be thrown into `setUp`

// Act - ideally, 1 line. Run the thing you want to test.

// Assert - Given the inputs from arrange, assert anything that should be true as a side-effect/output.

2. Why are the two datastore lines in this test? You don't appear to be retrieving from the datastore.

3. Why are you putting the EmbeddedEntity into the comment entity as a property? In the assert block, from what I can tell, you are retrieving the same property that you set above, meaning you're mostly just testing the test, rather than some underlying feature.

My concrete suggestions are to move or even copy any arrange logic into this test to show what you're testing. This may just be replacing `comment.getCommentText()` with `COMMENT_TEXT` so I more clearly see you're testing that input matches output.

From there, you've already run the function that you're claiming you want to test in the test name, so I would just assert that the object maintained the values you input. It's possible you're also testing another feature or set of features with this unit test, in which case I would recommend additional unit tests named to identify what you're testing, and then using the 3 block structure so I can more clearly follow what the `// Act` is that you're testing.

capstone/src/test/java/com/google/sps/objects/MemberResponseTest.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Test
+  public void fromEntityTest() {
+    Entity entity = new Entity("MemberResponse");
+    entity.setProperty("profilePic", PROFILE_PIC);
+    entity.setProperty("firstName", FIRST_NAME);
+    entity.setProperty("lastName", LAST_NAME);
+    entity.setProperty("userId", USER_ID);
+    MemberResponse returnedMemberResponse = MemberResponse.fromEntity(entity, false /*includeBadges*/);
```
g-asweet: I'd put a new line above this `fromEntity` call

g-asweet: Could you also add test cases for the includeBadges: true/false cases?

capstone/src/test/java/com/google/sps/objects/PostTest.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  @Test
+  public void removeLikeMultipleTest() {
+    post.addLike("test user 1");
+    post.addLike("test user 2");
+    HashSet<String> likes = post.getLikes();
+    assertEquals(likes.size(), 2);
+
+    post.removeLike("test user 1");
```
g-asweet: I'd put a new line under this line

g-asweet (dismissed):

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
anikabagga: changed in previous PR

capstone/src/main/java/com/google/sps/servlets/UpdateLikesServlet.java
anikabagga: changed in previous PR

capstone/src/main/java/com/google/sps/Objects/response/MemberResponse.java
anikabagga: fixed on previous PR

capstone/src/test/java/com/google/sps/objects/MemberResponseTest.java
anikabagga: Will add a test for the true case of includeBadges once I implement badges

capstone/src/main/java/com/google/sps/Objects/Comment.java
anikabagga: I agree that my current approach is simply checking if the comment value is the expected value, but doesn't check for an unique identifier (timestamp/commentId). Since the comment is an embedded entity, I can't generate an Id for the comment. 
Datastore allows embedded entities to have keys: "You can optionally associate a key with an embedded entity, but (unlike a full-fledged entity) the key is not required and, even if present, cannot be used to retrieve the entity." However, since I can't retrieve a comment by key this may pose a problem for editing comments, so we may need to change this to a regular entity. 

I'll keep this in mind, but will leave the equals check as is for now.

capstone/src/test/java/com/google/sps/objects/CommentTest.java
anikabagga: Looking back, I did this in the most roundabout way and basically just tested the test. I modified this to make a simple toEntity/fromEntity test to make sure the correct data was being stored. Thanks for explaining this!!

capstone/src/test/java/com/google/sps/objects/CommentTest.java
```diff
@@ -63,31 +65,11 @@ public void getUserTest() {
   }
 
   @Test
-  public void getCommentEntityTest() {
-    EmbeddedEntity entity = new EmbeddedEntity();
-    entity.setProperty("userId", USER_ID);
-    entity.setProperty("commentText", COMMENT_TEXT);
-    entity.setProperty("timestamp", TIMESTAMP);
-    Comment returnedComment = Comment.getCommentEntity(entity);
+  public void toAndFromCommentEntityTest() {
```
nicorod-google, line 68: I know Himani was having some issues with this test because some data needed to be stored in the DB to be formatted correctly, so double check this is working as intended and passing. If it isn't, you likely just need to follow what she did for the toFromEntityTest in Objects/UserTest.java

capstone/src/test/java/com/google/sps/objects/PostTest.java
```diff
@@ -157,7 +158,7 @@ public void removeLikeMultipleTest() {
   }
 
   @Test
-  public void getAndCreatePostEntityTest() {
+  public void toAndFromPostEntityTest() {
```
nicorod-google: I know Himani was having some issues with this test because some data needed to be stored in the DB to be formatted correctly, so double check this is working as intended and passing. If it isn't, you likely just need to follow what she did for the toFromEntityTest in Objects/UserTest.java

capstone/src/test/java/com/google/sps/objects/CommentTest.java
anikabagga: I checked again and my tests are working as intended. The correct values are being stored in toEntity and fromEntity. I'll ask her what wasn't passing just in case.

nicorod-google (approved):

PR #28, opened 2020-07-14
Group url

lxcyqx:
Load data on group specific data to page


capstone/src/main/java/com/google/sps/Objects/Challenge.java
```diff
@@ -79,7 +79,7 @@ public void addCompletedUser(String userId) {
    * @return boolean
    */
   public boolean getHasUserCompleted(String userId) {
-    if (this.usersCompleted == null) {
+    if (this.usersCompleted == null || this.usersCompleted == null) {
```
himaniyadav: What is this extra condition for?

capstone/src/main/java/com/google/sps/Objects/Group.java
```diff
@@ -13,8 +20,14 @@
   private final String headerImg;
   private final long groupId;
 
-  public Group(ArrayList<String> memberIds, ArrayList<Challenge> challenges, ArrayList<Post> posts, 
-      ArrayList<Option> options, String groupName, String headerImg, long groupId) {
+  public Group(
+      ArrayList<String> memberIds,
+      ArrayList<Challenge> challenges,
+      ArrayList<Post> posts,
+      ArrayList<Option> options,
```
himaniyadav: Change these to Long

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -5,6 +5,12 @@ let maxVotes;
 let topChallenge = null;
 const NO_CHALLENGES =
   "No current challenges. Submit a suggestion in the poll and mark checkbox for challenge to be posted. The challenge will be updated weekly based on top voted poll option.";
+let groupId;
+
+function getGroupId() {
+  groupId = window.location.search.substring(1).split("=")[1];
+  console.log(groupId);
```
himaniyadav: Can probably remove the log statements

capstone/src/test/java/com/google/sps/servlets/PollServletTest.java
```diff
@@ -109,6 +141,8 @@ public void tearDown() {
 
   @Test
   public void doGet_userLoggedIn() throws IOException {
+    this.addOptionsToGroup();
+    when(Group.getGroupEntity(mockRequest, mockResponse, datastore)).thenReturn(GROUP_ENTITY);
     pollServlet.doGet(mockRequest, mockResponse);
```
g-asweet, line 136: Can you put new lines before and after the `.doGet` call so you get a three block

```java
// Arrange

// Act

// Assert
```

structure to your tests?

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -25,22 +23,54 @@
...
+
+    for (Long challengeId : challengeIds) {
+      // get newest challenge
+      Entity entity = ServletHelper.getEntityFromId(response, challengeId, datastore, "Challenge");
+      Long timestamp = (Long) entity.getProperty("timestamp");
+      if (timestamp > newestTimestamp) {
+        newestTimestamp = timestamp;
+        newestChallengeEntity = entity;
+      }
+    }
```
g-asweet: Looking at `updateChallengesList` it looks like you're already maintaining an order to chellengeIds based on recency. Can this order be used to guarantee that any call to `getMostRecentChallenge` only checks AT MOST all challenges which have been invalidated since the last time `getMostRecentChallenge` was called?

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -54,6 +84,24 @@ public void doPost(String userId, HttpServletRequest request, HttpServletRespons
...
+      HttpServletResponse response,
+      DatastoreService datastore,
+      Entity challengeEntity)
+      throws IOException {
+    Entity groupEntity = Group.getGroupEntity(request, response, datastore);
+    List<Long> challenges =
+        (groupEntity.getProperty("challenges") == null)
+            ? new ArrayList<Long>()
+            : (List<Long>) groupEntity.getProperty("challenges");
+    challenges.add(challengeEntity.getKey().getId());
```
g-asweet, line 108: [This is comment without any work required.] This order is only strictly maintained by the fact that all challenges CURRENTLY expire after the same amount of time, otherwise you may want to consider sorting by the `timestamp` property, especially since order-of-creation is easily checked by comparing IDs if you ever need to recreate that list.

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -50,34 +47,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    for (Long optionId : options) {
+      Entity optionEntity = ServletHelper.getEntityFromId(response, optionId, datastore, "Option");
+      List<String> votes = (ArrayList<String>) optionEntity.getProperty("votes");
       int numVotes = (votes != null) ? votes.size() : 0;
       if (numVotes > maxVotes) {
         maxVotes = numVotes;
-        maxVotedId = id;
+        maxVotedOptionId = optionId;
       }
     }
```
g-asweet, line 80: Maybe better for a follow-up CL (or leaving this as a mental TODO which never needs to get done):

Currently, you're having to do all of this logic every time anybody calls doPost. But one nice property of this calculation is it wouldn't change if nobody added or removed any votes or poll options between `getMaxVotedOption` calls on a given poll.

Could you redesign this servlet/database to allow `O(1)` retrieval in cases where no votes were changes and no poll options were added or removed?

g-asweet (dismissed):

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
```diff
@@ -25,22 +23,54 @@
...
+    }
+    if (newestChallengeEntity == null) {
+      return null;
+    }
+    // If the newest challenge's due date already passed
+    if ((long) newestChallengeEntity.getProperty("dueDate") < System.currentTimeMillis()) {
+      return null;
+    }
+    newestChallenge = Challenge.fromEntity(newestChallengeEntity);
+    return newestChallenge;
```
nicorod-google: Since newestChallenge isn't used until the end, you can just return Challenge.fromEntity(newestChallengeEntity) and get rid of newestChallenge entirely

capstone/src/main/webapp/group.html
```diff
@@ -17,6 +17,9 @@
     <script src="scripts/poll.js"></script>
     <script src="scripts/members.js"></script>
     <script src="https://unpkg.com/ionicons@5.1.2/dist/ionicons.js"></script>
+    <script>
+      window.onload = getGroupId();
+ </script>
```
nicorod-google: Could this be added to the init in posts.js instead? Or perhaps create a new group.js file and move the init from posts.js into there?

capstone/src/test/java/com/google/sps/servlets/PollServletTest.java
```diff
@@ -48,8 +50,12 @@
   private static final String USER_EMAIL = "test@test.com";
   private static final String USER_ID = "test";
   private static final String NEW_OPTION = "Do a 5k";
+  private static final long NEW_OPTION_ID = 2;
   private static final List<String> OPTION_TEXT =
       new ArrayList<String>(Arrays.asList("Run", "Jog", "Climb a tree", "Bungee jump"));
+  private static final String GROUP_NAME = "Runners Club";
+  private static final String GROUP_ID = "1";
+  //  private static final Entity GROUP_ENTITY = new Entity("Group");
```
nicorod-google: nit: remove

capstone/src/test/java/com/google/sps/servlets/PollServletTest.java
```diff
@@ -66,28 +72,27 @@
...
-      option.add(createOption(text));
-    }
-    datastore.put(option.build());
+    //    // Add test data
+    //    ImmutableList.Builder<Entity> option = ImmutableList.builder();
+    //    for (String text : OPTION_TEXT) {
+    //      option.add(createOption(text));
+    //    }
+    //    datastore.put(option.build());
+    //    datastore.put(createGroup(USER_ID, GROUP_NAME));
```
nicorod-google: nit: remove

capstone/src/test/java/com/google/sps/servlets/PollServletTest.java
```diff
@@ -107,15 +112,50 @@ public void tearDown() {
...
+    datastore.put(groupEntity);
+  }
+
   @Test
   public void doGet_userLoggedIn() throws IOException {
+    Entity groupEntity = this.createGroup(USER_ID, GROUP_NAME);
+    datastore.put(groupEntity);
+    System.out.println(groupEntity);
+    System.out.println("in do get user logged in --------------");
+    //    this.addOptionsToGroup();
```
nicorod-google: nit: remove

capstone/src/test/java/com/google/sps/servlets/PollServletTest.java
```diff
@@ -107,15 +112,50 @@ public void tearDown() {
...
+    when(mockRequest.getParameter("groupId")).thenReturn(GROUP_ID);
+    doReturn(groupEntity).when(pollServlet).getGroupEntity(mockRequest, mockResponse, datastore);
+    System.out.println("after when --------");
+    //        .thenReturn(new Entity("Group"));
```
nicorod-google: nit: remove

capstone/src/test/java/com/google/sps/servlets/PollServletTest.java
```diff
@@ -107,15 +112,50 @@ public void tearDown() {
...
     pollServlet.doGet(mockRequest, mockResponse);
     String response = responseWriter.toString();
-
+    System.out.println(response);
```
nicorod-google: nit: remove

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
lxcyqx: changed it so that now I get the challengeId that is at the end of the arraylist since that should be the most recent one

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
```diff
@@ -50,34 +47,36 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    int maxVotes = 0;
-    Iterable<Entity> resultsIterable = results.asIterable();
-    // Set maxVotedId to be first one in iterable of Entities
-    if (Iterables.size(resultsIterable) > 0) {
-      maxVotedId = results.asIterable().iterator().next().getKey().getId();
+  private List<Long> getOptionsList(
+      HttpServletRequest request, HttpServletResponse response, DatastoreService datastore)
+      throws IOException {
+    Group group = new Group(null, null, null, null, "", "", 0);
+    Entity entity = group.getGroupEntity(request, response, datastore);
```
nicorod-google: Since the getGroupEntity method is static, can you not just call Group.getGroupEntity instead of creating this dummy group that isn't used?

nicorod-google (approved):

capstone/src/main/java/com/google/sps/servlets/ChallengeServlet.java
lxcyqx: I think the list is ordered by when it is added into the database though

capstone/src/main/java/com/google/sps/servlets/DeletePollOptionServlet.java
lxcyqx: I think this servlet is only called when the challenge needs to be updated, so this is handled on the javascript side. I check whether the challenge has passed and if it has, then I call this servlet to delete the top poll option and update the challenge

PR #32, opened 2020-07-16
Add Members to Group

anikabagga:
- add members to group by email 
- show only members in specific group 
- posts specific to group 

Screenshot 2020-07-16 at 9 29 10 AM


capstone/src/test/java/com/google/sps/servlets/GroupMemberServletTest.java
```diff
@@ -114,7 +114,7 @@ public void doPost_userNotLoggedIn() throws IOException, EntityNotFoundException
     assertTrue(response.contains("Oops an error happened!"));
   }
 
-  @Test
+  //@Test
```
g-asweet: Can you restore these tests and make sure they're working?

capstone/src/test/java/com/google/sps/servlets/GroupPostDataServletTest.java
```diff
@@ -105,15 +105,15 @@ private void populateDatabase(DatastoreService datastore) {
     datastore.put(POST_1.toEntity());
   }
 
-  @Test
+  //@Test
```
g-asweet: Can you restore these tests and make sure they're working?

capstone/src/test/java/com/google/sps/servlets/GroupPostDataServletTest.java
```diff
@@ -105,15 +105,15 @@ private void populateDatabase(DatastoreService datastore) {
...
   public void doGet_userLoggedIn() throws Exception {
     groupPostDataServlet.doGet(mockRequest, mockResponse);
     String response = responseWriter.toString();
 
     assertTrue(response.contains(POST_1.getPostText()));
   }
 
-  @Test
+  //@Test
```
g-asweet: Can you restore these tests and make sure they're working?

capstone/src/test/java/com/google/sps/servlets/GroupMemberServletTest.java
```diff
@@ -104,7 +104,7 @@ private void populateDatabase(DatastoreService datastore) {
     datastore.put(group1);
   }
 
-  @Test
+  //@Test
```
g-asweet: Can you restore these tests and make sure they're working?

capstone/src/test/java/com/google/sps/servlets/AllGroupMembersServletTest.java
```diff
@@ -103,12 +109,25 @@ public void tearDown() {
...
+    Entity groupEntity = new Entity("Group", groupId);
+    groupEntity.setProperty("groupId", groupId);
+    groupEntity.setProperty("groupName", GROUP_NAME);
+    groupEntity.setProperty("headerImg", HEADER_IMAGE);
+    groupEntity.setProperty("memberIds", new ArrayList<String>(Arrays.asList(USER_1_ID, USER_2_ID)));
+    groupEntity.setProperty("challenges", new ArrayList<Challenge>());
+    return groupEntity;
+  }
+
+  //@Test
```
g-asweet: Can you restore these tests and make sure they're working?

capstone/src/main/webapp/styles/members.css
```diff
@@ -120,8 +123,59 @@
   margin: 0 auto;
 }
 
+<<<<<<< HEAD
```
g-asweet: You seem to have some unresolved merge conflict issues

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,21 +1,18 @@
...
-      allPostsList.innerHTML = "";
-      for (let i = 0; i < posts.length; i++) {
-        allPostsList.appendChild(
-          createSinglePost(posts[i], postResponse["likedPosts"])
-        );
-      }
-      addCommentInputListener();
-      addLikeButtonListener();
-    });
+  console.log(groupId);
```
g-asweet: Can you remove this console.log?

capstone/src/main/webapp/scripts/post.js
```diff
@@ -1,21 +1,18 @@
...
+  fetch(`/group-post?groupId=${groupId}`).then(response => response.json()).then((postsResp) => {
+    postResponse = postsResp;
```
g-asweet: You shouldn't need this line, you should be able to change `(postsResp) =>` to `(postResponse) =>`

capstone/src/main/webapp/scripts/post.js
```diff
@@ -206,12 +203,13 @@ function createCommentBox(post) {
...
 // Gets URL for uploaded image
 function fetchBlobstoreUrlAndShowForm() {
-  fetch("/post-image")
+  fetch(`/post-image?groupId=${groupId}`)
     .then(response => {
       return response.text();
     })
     .then(imageUploadUrl => {
       const messageForm = document.getElementById("post-form");
+      console.log(imageUploadUrl);
```
g-asweet: Can you remove this `console.log`?

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -26,24 +33,63 @@
...
+        group.setProperty("memberIds", members);
+        datastore.put(group);
+        memResponse += "User added to group.";
+        addGroupToUser(response, datastore, groupId, memberEntity);
+      }
     }
+    return memResponse;
+  }
+
+  private void addGroupToUser(HttpServletResponse response, DatastoreService datastore, Long groupId, Entity memberEntity) {
```
g-asweet: Can you restructure this to return true if a groupId is successfully added and false if one is already found?

then you can just do

```java
if (maybeAddGroupToUser(...)) {
  memResponse += "User added to group."
} else {
  memResponse += "User is already in group."
}
```

in doPost_helper and I won't lose sleep over whether addGroupToUser is ever going to accidentally be used to add a second instance of the groupId to the the user 😅

g-asweet (dismissed):

capstone/src/main/java/com/google/sps/servlets/AllGroupMembersServlet.java
```diff
@@ -17,26 +17,59 @@
...
+
+  private Entity getUserFromId(
+    HttpServletResponse response,String userId, DatastoreService datastore)  throws IOException {
+      try {
+        return datastore.get(KeyFactory.createKey("User", userId));
+      } catch (EntityNotFoundException e) {
+        errorHandler.sendError(response, "Group does not exist.");
+        return null;
+      }
+  }
```
nicorod-google: You can use the ServletHelper.getEntityFromId method instead of needing to create these methods for each entity

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -45,43 +47,94 @@
...
+  }
+
+  private Entity getPostFromId(
+    HttpServletResponse response, Long postId, DatastoreService datastore) throws IOException {
+    try {
+      return datastore.get(KeyFactory.createKey("Post", postId));
+    } catch (EntityNotFoundException e) {
+      ErrorHandler.sendError(response, "Post does not exist.");
+      return null;
+    }
```
nicorod-google: You can use the ServletHelper.getEntityFromId method instead of needing to create these methods for each entity

capstone/src/main/java/com/google/sps/servlets/PostImageServlet.java
```diff
@@ -21,8 +21,12 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    
+    Long groupId = Long.parseLong(request.getParameter("groupId"));
+
     BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
-    String uploadUrl = blobstoreService.createUploadUrl("/group-post");
+    String url = "/group-post?groupId="+groupId;
```
nicorod-google: nit: add spaces before and after the +

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -26,49 +34,71 @@
...
-    
+
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    Entity group = this.getGroupFromId(response, groupId, datastore);
-    if (group != null && !userId.equals("")) {
+    Entity memberEntity = getMemberEntity(email, response, datastore);
+    Entity group = ServletHelper.getEntityFromId(response, groupId, datastore, "Group");
+    if (group == null) return;
+
+    String memResponse = doPost_helper(response, datastore, groupId, memberEntity, group);
```
nicorod-google: It doesn't look like you are actually using memResponse for anything right now? I want to understand what your intention is here?

capstone/src/main/java/com/google/sps/servlets/ServletHelper.java
```diff
@@ -63,4 +68,24 @@ public static Entity getEntityFromId(
...
+  public static Entity getUserFromId(
+      HttpServletResponse response, String id, DatastoreService datastore)
+      throws IOException {
+    try {
+      return datastore.get(KeyFactory.createKey("User", id));
+    } catch (EntityNotFoundException e) {
+      ErrorHandler.sendError(response, "Cannot get entity from datastore");
+      return null;
+    }
+  }
```
nicorod-google, line 90: The only difference that I see between getUserFromId and getEntityFromId is that one uses a string id and the other uses a long. Rather than write a specific one for a User entity, I recommend either 
a) Converting the string id to a long and call getEntityFromId
b) Create an overloaded getEntityFromId method that takes in a string for the id but then calls the other getEntityFromId and converts the string id to a long.
If you need to use getUserFromId many times I recommend option b, otherwise a is fine as well

nicorod-google (changes requested):

capstone/src/test/java/com/google/sps/objects/PostTest.java
```diff
@@ -177,12 +177,7 @@ public void removeLikeMultipleTest() {
   @Test
   public void toAndFromPostEntityTest() throws Exception {
     Entity entity = post.toEntity();
-    datastore.put(entity);
-
-    Key postKey = KeyFactory.createKey("Post", POST_ID);
-    Entity retreivedPost = datastore.get(postKey);
-
-    Post returnedPost = Post.fromEntity(retreivedPost);
+    Post returnedPost = Post.fromEntity(entity);
```
himaniyadav, line 180: Not if this comment is relevant anymore, but I just pulled from master and ran the old version of this test, and it was failing. Changing line 64 of post.toEntity to `Entity entity = new Entity("Post", this.postId);` would fix the test but caused issues with being unable to add posts to a group, so probably not the right solution. I guess as long as the test passes like this then you're good!

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -45,43 +48,65 @@
...
+  }
+
+  private void addPostToGroup(HttpServletResponse response, Long groupId, DatastoreService datastore, Entity postEntity) throws IOException {
+    Entity groupEntity = ServletHelper.getEntityFromId(response, groupId, datastore, "Group");
+    ArrayList<Long> postIds = 
+        (ArrayList<Long>) groupEntity.getProperty("posts");
+    if (postIds == null) {
+      postIds = new ArrayList<Long>();
+    }
+    Post storedPost = Post.fromEntity(postEntity);
```
lxcyqx: Would it be easier to just do postEntity.getKey().getId()

capstone/src/main/java/com/google/sps/servlets/ServletHelper.java
lxcyqx: I think the userId is larger than the maximum allowed for a long so we might need to have it be what Anika has right now

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
anikabagga: I was using it for my own documentation, which could eventually be sent as json and displayed to the user to let them know if the user was added, etc.

nicorod-google: Since you aren't using it atm then just call doPost_helper without assigning it to a variable

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -26,68 +34,92 @@
...
+  }
+
+  // If no account for email, return 'This email doesn't have an account'
+  // If user already in group, return 'User already in group'
+  // If user not it group, add user and return 'User added to group'
+  private String doPost_helper(HttpServletResponse response, DatastoreService datastore, Long groupId, Entity memberEntity, Entity group) {
```
nicorod-google: doPost_helper should be all camel case. Also, please rename the method since the name isn't very descriptive of what it is doing. possibly something like maybeAddUserToGroup

capstone/src/main/java/com/google/sps/servlets/GroupPostDataServlet.java
```diff
@@ -45,43 +48,63 @@
 
   @Override
   public void doGet(String userId, HttpServletRequest request, HttpServletResponse response) throws IOException {
-
-    Query query = new Query("Post").addSort("timestamp", SortDirection.DESCENDING);
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    PreparedQuery results = datastore.prepare(query);
+    Long groupId = Long.parseLong(request.getParameter("groupId"));
+    Entity groupEntity = ServletHelper.getEntityFromId(response, groupId, datastore, "Group");
+    System.out.println("THE GROUP: " + groupEntity.getProperty("groupName"));
```
nicorod-google: nit: remove

nicorod-google (approved):

capstone/src/main/java/com/google/sps/servlets/ServletHelper.java
nicorod-google: Thanks for the explanation!

capstone/src/main/java/com/google/sps/servlets/GroupMemberServlet.java
```diff
@@ -26,68 +34,85 @@
...
+
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    Entity group = this.getGroupFromId(response, groupId, datastore);
-    if (group != null && !userId.equals("")) {
+    Entity memberEntity = getMemberEntity(email, response, datastore);
+    if (memberEntity == null) return;
+    Entity group = ServletHelper.getEntityFromId(response, groupId, datastore, "Group");
+    if (group == null) return;
+
+    doPost_maybeAddUserToGroup(response, datastore, groupId, memberEntity, group);
```
nicorod-google, line 47: nit: remove the doPost_

nicorod-google (approved):

PR #33, opened 2020-07-17
Make email value read-only

himaniyadav:
While we initially mocked in the Figma that users could change their email from the edit profile modal, we realized it didn't really make sense. Even if they changed their email in the profile, they would still have to log in with the same google account through the Users API. Additionally, since you can search and add users to group by email, we did not want that to be an editable value.
- Still in the UI as a read only form value so that users can still see it along with the rest of their profile data
image

### DIFF
--- a/capstone/src/test/java/com/google/sps/objects/TimeTest.java
+++ b/capstone/src/test/java/com/google/sps/objects/TimeTest.java
@@ -17,7 +17,7 @@ import com.google.sps.Objects.Time;
 public class TimeTest {
 
   @Spy private Time mockedTime;
-  private static long EXPECTED_DATE = 1595044799000L;
+  private static long EXPECTED_DATE = 1595030399000L;
 
   @Before
   public void setUp() throws IOException {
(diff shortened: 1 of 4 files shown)


capstone/src/test/java/com/google/sps/objects/TimeTest.java
```diff
@@ -17,7 +17,7 @@
 public class TimeTest {
 
   @Spy private Time mockedTime;
-  private static long EXPECTED_DATE = 1595044799000L;
+  private static long EXPECTED_DATE = 1595030399000L;
```
nicorod-google: I think at least someone else has a different value for the expected date so these two keep changing on every PR lol. I'm just gonna ask you to keep the old one so it doesn't always change

nicorod-google (approved):

capstone/src/test/java/com/google/sps/objects/TimeTest.java
himaniyadav: Yeah we're not really sure why it keeps changing, but we keep have to switching so tests don't fail. I'll switch it back though!

PR #34, opened 2020-07-17
Join group

lxcyqx:
- Allow users to join group if they are not already in it
- Redirect user to log in page if they are not logged in / do not have an account
- Make group banner name specific to group
- Improve group page responsiveness to window resizing
- Add tests for JoinGroupServlet

Screen Shot 2020-07-17 at 7 30 36 PM
Screen Shot 2020-07-17 at 7 46 49 PM

### DIFF
--- a/capstone/src/main/java/com/google/sps/servlets/AuthenticatedServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/AuthenticatedServlet.java
@@ -21,6 +21,8 @@ public abstract class AuthenticatedServlet extends HttpServlet {
       this.doGet(userId, request, response);
     } else {
       ErrorHandler.sendError(response, "User is not logged in.");
+      String loginUrl = userService.createLoginURL("/");
+      response.sendRedirect(loginUrl);
     }
   }
 
--- a/capstone/src/main/java/com/google/sps/servlets/AuthenticatedServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/AuthenticatedServlet.java
@@ -9,8 +9,6 @@ import javax.servlet.http.HttpServletResponse;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
 
-import error.ErrorHandler;
-
 public abstract class AuthenticatedServlet extends HttpServlet {
 
   @Override
@@ -20,9 +18,7 @@ public abstract class AuthenticatedServlet extends HttpServlet {
       String userId = userService.getCurrentUser().getUserId();
       this.doGet(userId, request, response);
     } else {
-      ErrorHandler.sendError(response, "User is not logged in.");
-      String loginUrl = userService.createLoginURL("/");
-      response.sendRedirect(loginUrl);
+      this.redirectToLogin(userService, response);
     }
   }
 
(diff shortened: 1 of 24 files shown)


capstone/src/main/java/com/google/sps/servlets/AuthenticatedServlet.java
```diff
@@ -21,6 +21,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       this.doGet(userId, request, response);
     } else {
       ErrorHandler.sendError(response, "User is not logged in.");
```
nicorod-google: If you are sending a response do we still need the send the error?

capstone/src/main/webapp/scripts/group.js
```diff
@@ -1,8 +1,65 @@
+let group_id;
```
nicorod-google: nit: this should be camel case

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -5,8 +5,8 @@ let maxVotes;
 let topChallenge = null;
 const NO_CHALLENGES =
   "No current challenges. Submit a suggestion in the poll and mark checkbox for challenge to be posted. The challenge will be updated weekly based on top voted poll option.";
-let groupId;
 
+let groupId;
 function getGroupId() {
   groupId = window.location.search.substring(1).split("=")[1];
 }
```
nicorod-google: I think we can remove the group id related logic in this file. Since we are not namespacing the JS files, we're able to reference other files' functions without needing to import them. I see you added a similar getId() method in group.js, which we can reference instead of this one (I would also change the name to the same as this one so it makes more sense when you use it outside of group.js). Additionally, since group_id is a global variable, you'll similarly be able to use that across other files (though similarly, change the name to groupId. Since that's the name in poll.js you'll just need to change the references in group.js).

capstone/src/main/webapp/scripts/poll.js
```diff
@@ -265,3 +265,5 @@ function markChallenge(id, checked) {
     method: "POST"
   });
 }
+
+getGroupId();
```
nicorod-google: I don't think this will be needed after the above changes are made

capstone/src/main/webapp/scripts/group.js
```diff
@@ -1,8 +1,65 @@
...
-  getGroupId();
   loadPosts();
   getPollOptions();
   fetchBlobstoreUrlAndShowForm();
   loadMembers();
 }
+
+function getId() {
+  group_id = window.location.search.substring(1).split("=")[1];
+}
```
nicorod-google: refer to my comment in poll.js about this function and group_id

nicorod-google (changes requested):

capstone/src/test/java/com/google/sps/servlets/JoinGroupServletTest.java
```diff
@@ -0,0 +1,203 @@
...
+    userEntity.setProperty("interests", null);
+    return userEntity;
+  }
+
+  @Test
+  public void doGet_userIsMember() throws IOException {
+    List<String> members = new ArrayList<String>();
+    members.add(USER_ID);
+    Entity groupEntity = this.createGroup(USER_ID, GROUP_NAME, members);
+    datastore.put(groupEntity);
```
himaniyadav, line 114: I think this might pass anyway because how you're doing the check in JoinGroupServlet, but shouldn't you also add the groupId to the user's list of groups?

himaniyadav: Wait, I see that you tested for that later on, so maybe it's not necessary.

nicorod-google (approved):

capstone/src/test/java/com/google/sps/servlets/JoinGroupServletTest.java
lxcyqx: Yeah, in my getIsMember() method on JoinGroupServlet, I only check whether the group's list of members contains the member. I do update the user's list of groups, but that's not part of my check for whether the user is a member, so that is why I have two separate tests

PR #36, opened 2020-07-21
Database storing

lxcyqx:
Store test name data in database through web.xml
Save instance of SearchPredictor in file and make sure it is stateful (as long as the server is running, all users will be accessing this one instance of SearchPredictor)


capstone/src/main/webapp/group.html
```diff
@@ -72,11 +72,12 @@ <h4 class="align-vertical">Members</h4>
           <span id="add-close" onclick="addModalListeners('add-member-modal', 'add-close')">&times;</span>
           <h2>Add Group Member</h2>
           <form id="add-member" onsubmit="addMember()" method="POST">
-            <label for="member-email-input">Enter an email</label>
+            <label for="member-email-input">Enter email</label>
             <input type="text" class="input-field" name="email-input" id="email-input">
             <p id=memberUpdate></p>
             <button class="add-member-btn">Add</button>
           </form>
+          <button class="add-member-btn" onclick="loadTestData()">Load Test Data</button>
```
himaniyadav: I think one thing to note is that pressing this button would overwrite Anika's data (Users w/ longitude and latitude values) if hers is already loaded in the db. An alternative way of handling it could be to have your NameTestData servlet `load-on-startup` (specified in the `web.xml` file). [This link](https://beginnersbook.com/2017/07/load-on-startup-tag-in-web-xml-file-with-example/) goes into how you can specify the order of the servlets, so as long as yours comes before Anika's, there shouldn't be an issue with her data being overwritten.

This will also help avoid the case of someone accidentally pressing the button twice and loading duplicate data (wouldn't actually be duplicate thanks to the user ids, would just be overwritten, but servlet would still go through the unnecessary work of putting each user in the datastore).

capstone/src/main/java/com/google/sps/search/servlets/NameTestData.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.sps.Objects.User;
+import com.google.sps.search.CsvReader;
+
+@WebServlet("/name-test-data")
+public class NameTestData extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) {
```
himaniyadav: If you do end up going the load-on-startup route, this would be changed to `init()`

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.Set;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+
+public class SearchPredictor implements Serializable {
```
g-asweet, line 18: Interesting, I didn't expect you to serialize but this does allow you not have to recreate this every time you start the same server (or in theory, a new server if you could load from the serialized data on another server).

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
```diff
@@ -0,0 +1,59 @@
...
+      names.add(name);
+    }
+    return names;
+  }
+
+  private void populateTrie() {
+    for (String fullName : names) {
+      String[] split = fullName.split(" ");
+      String firstName = split[0];
+      String lastName = split[1];
```
g-asweet: I'd add a comment above this acknowledging that firstname/lastname is not as robust for i18n (internationalization) and other considerations.

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private List<String> getNamesFromDb() {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("User");
+    PreparedQuery pq = datastore.prepare(query);
+    List<String> names = new ArrayList<String>();
+    for (Entity userEntity : pq.asIterable()) {
+      String firstName = (String) userEntity.getProperty("firstName");
+      String lastName = (String) userEntity.getProperty("lastName");
+      String name = firstName + " " + lastName;
```
g-asweet: I'm wondering if instead of combining them by a " " and then separating it on what could be a DIFFERENT space if you want to instead either not combine them to avoid having to split them, or use a unique separator so you can guarantee that you're using what the user calls their firstName and lastName ("anne marie rose")

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -0,0 +1,35 @@
...
+  private boolean isEnd;
+
+  public Trie() {
+    children = new HashMap<String, Trie>();
+    isEnd = false;
+  }
+
+  public void insert(String name, String fullName) {
+    if (name.contentEquals("")) {
+      isEnd = true;
```
g-asweet: `isEnd` makes enough sense right now, but my guess is you can just replace `isEnd` with a set of full names, and if the set is non-empty, we know it's a valid end.

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
```diff
@@ -0,0 +1,78 @@
...
+    try {
+      FileInputStream fileInput = new FileInputStream(new File(trieFile));
+      ObjectInputStream objectInput = new ObjectInputStream(fileInput);
+      searchPredictor = (SearchPredictor) objectInput.readObject();
+
+      fileInput.close();
+      objectInput.close();
+      return;
+    } catch (FileNotFoundException e) {
+      System.err.println("File does not exist");
```
g-asweet, line 49: How does this work for a new server? Are you just going to guarantee that any server already has the `data/trie` file instead of generating their own?

capstone/src/main/java/com/google/sps/search/servlets/NameTestDataServlet.java
```diff
@@ -0,0 +1,31 @@
...
+  @Override
+  public void init() {
+    CsvReader reader = new CsvReader();
+    List<String> userInfo = reader.parseCorpus("../../data/user_data.csv");
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    for (String str : userInfo) {
+      String[] split = str.split(" ");
+      String firstName = split[0];
+      String lastName = split[1];
+      String userId = firstName + "_" + lastName;
```
g-asweet: This doesn't guarantee uniqueness if you have multiple users with the same first/last name

g-asweet (dismissed):

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
lxcyqx: I have a file in the data folder and each time the server is initialized, it uses this file

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
lxcyqx: added in other pr

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
lxcyqx: Hmm yeah I considered this too. I was thinking of having a list of first names and a list of last names, but then I wouldn't be able to link the two. I'm thinking I can create a full name object that has a first name and last name field and then have a list of full name objects. However maybe having a unique separator may be better in terms of space

capstone/src/main/java/com/google/sps/search/CsvReader.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public List<String> parseCorpus(String file) {
+    List<String> names = new ArrayList<String>();
+    Scanner in = null;
+    try {
+      in = new Scanner(new File(file));
+    } catch (FileNotFoundException e) {
+      System.err.println("File does not exist.");
+      return null;
+    }
```
nicorod-google, line 21: You should make sure to close your file in the catch otherwise you may have unexpected consequences

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
```diff
@@ -0,0 +1,78 @@
...
+
+import com.google.sps.search.SearchPredictor;
+import com.google.sps.servlets.AuthenticatedServlet;
+import com.google.sps.servlets.ServletHelper;
+
+@WebServlet("/name-data")
+public class NameDataServlet extends AuthenticatedServlet {
+
+  SearchPredictor searchPredictor;
+  String trieFile = "../../data/trie";
```
nicorod-google: nit: Make this a constant

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
```diff
@@ -0,0 +1,78 @@
...
+      fileInput.close();
+      objectInput.close();
+      return;
+    } catch (FileNotFoundException e) {
+      System.err.println("File does not exist");
+    } catch (IOException e) {
+      System.err.println("Cannot read from file");
+    } catch (ClassNotFoundException e) {
+      System.err.println("Class not found");
+    }
```
nicorod-google, line 54: Make sure to close the files in the catch blocks

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
```diff
@@ -0,0 +1,78 @@
...
+  public void init(ServletConfig config) throws ServletException {
+    super.init(config);
+    try {
+      FileInputStream fileInput = new FileInputStream(new File(trieFile));
+      ObjectInputStream objectInput = new ObjectInputStream(fileInput);
+      searchPredictor = (SearchPredictor) objectInput.readObject();
+
+      fileInput.close();
+      objectInput.close();
+      return;
```
nicorod-google: If you return here, you don't get to line 45: 
`searchPredicctor = new SearchPredictor();`
I don't think that's intended?

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
```diff
@@ -0,0 +1,78 @@
...
+      objectOutputStream.writeObject(searchPredictor);
+
+      objectOutputStream.close();
+      fileOutputStream.close();
+      System.out.println("successfully written to file");
+    } catch (FileNotFoundException e1) {
+      System.err.println("File does not exist");
+    } catch (IOException e) {
+      System.err.println("Cannot write to file");
+    }
```
nicorod-google, line 80: nit: close the files here too

capstone/src/main/java/com/google/sps/search/servlets/NameTestDataServlet.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.sps.Objects.User;
+import com.google.sps.search.CsvReader;
+
+@WebServlet("/name-test-data")
+public class NameTestDataServlet extends HttpServlet {
+
+  @Override
+  public void init() {
+    CsvReader reader = new CsvReader();
+    List<String> userInfo = reader.parseCorpus("../../data/user_data.csv");
```
nicorod-google, line 19: parseCorpus returns null if there was an issue reading the data, you should make sure that userInfo is not null before continuing

capstone/src/main/java/com/google/sps/search/servlets/NameTestDataServlet.java
nicorod-google: +1 Since this is just test data, you don't need to do anything elaborate, maybe just a simple counter that increases with each name.

Also as an FYI in case you didn't know, while you're doing simple concatenation here and therefore not an issue, you'll typically want to combine strings using a StringBuilder instead of concatenating strings because it is inefficient:
https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java
https://dzone.com/articles/string-concatenation-performacne-improvement-in-ja
https://javapapers.com/java/java-string-vs-stringbuilder-vs-stringbuffer-concatenation-performance-micro-benchmark/#:~:text=Difference%20between%20String%2C%20StringBuilder%2C%20StringBuffer&text=Using%20'%2B'%20operator%20and%20String,concat%20are%20not%20the%20same.&text=%2B%20operator%20uses%20StringBuilder%20internally%20to,as%20below%20by%20Java%20compiler.

capstone/src/test/java/com/google/sps/objects/TimeTest.java
```diff
@@ -17,7 +17,7 @@
 public class TimeTest {
 
   @Spy private Time mockedTime;
-  private static long EXPECTED_DATE = 1595894399000L;
+  private static long EXPECTED_DATE = 1595908799000L;
```
nicorod-google, line 20: Does this need to change or can you keep it to the old value? This keeps changing in PRs and it's possibly because y'all have different versions of this that keep getting uploaded

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
nicorod-google: Better yet, if you don't need to return in the try block and subsequently don't need to in the catch blocks, just close the files once after the try-catch block

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
nicorod-google: Same comment: Better yet, you can just close the files once after the try-catch block

capstone/src/main/java/com/google/sps/search/servlets/NameDataServlet.java
lxcyqx: I initialize searchPredictor because it there's nothing to read it'll initialize the new object, which is what happens whenever the server starts

lxcyqx: I've moved it above the try and got rid of the return

nicorod-google (approved):

capstone/src/test/java/com/google/sps/objects/TimeTest.java
lxcyqx: I think this has something to do with whether this project is run on Cloud Shell or my personal computer. It always switches between two different values, where Cloud Shell is always one number and my personal computer is another value

PR #38, opened 2020-07-22
Trie methods

lxcyqx:
- Implement trie insert and search
- Ability to search for full name
- Connect search with backend
- Added full name field for querying. The full name field is all capital because datastore does not allow for case insensitive querying.
- Add member to group upon clicking plus icon
- Display results from search upon pressing enter, clicking on name, or pressing enter after cycling through suggestions using arrow keys
- If user clicks on one of the suggested names, the results will list users with matching names first
- Update tries when new user is added to database
- Changed map in Trie to be from Character to Trie instead of String to Trie
- fix tests

Screen Shot 2020-07-23 at 11 52 33 AM


capstone/src/main/java/com/google/sps/search/SearchPredictor.java
```diff
@@ -49,8 +50,12 @@ private void populateTrie() {
...
   }
 
-  public Set<String> suggest() {
-    return null;
+  // incomplete
+  public Set<String> suggest(String phrase) {
+    Set<String> names = new HashSet<String>();
+    names.addAll(firstNameTrie.searchWithPrefix(phrase, phrase));
+    names.addAll(lastNameTrie.searchWithPrefix(phrase, phrase));
+    return names;
```
g-asweet: Perfect mvp!

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -1,35 +1,122 @@
...
   public void insert(String name, String fullName) {
     if (name.contentEquals("")) {
-      isEnd = true;
-      // add in the full name here
+      // indicates the end of first name or last name
+      this.isName = true;
+      // insert full name node as child
+      Trie fullNameNode = new Trie();
+      fullNameNode.isFullName = true;
+      this.children.put(fullName, fullNameNode);
```
g-asweet: I see what you're doing, but this seems like an odd overloading of existing features.

I would suggest you keep the trie->child relation ship always 1 lowercase character. This is the kind of feature that breaks when a user has a 1 letter first or last name in the DB or something odd.

Instead, I would remove `isFullName` and replace it with `Set<String> fullNames` so you can just check if the set is empty, or get all unique full names.

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -1,35 +1,122 @@
...
+    /* If current node is the end of the a first name or last name, look through its map of children
+     * and add the full name nodes to set of names to be returned. */
+    if (this.isName) {
+      for (String child : possibilities.keySet()) {
+        StringBuilder sb = new StringBuilder();
+        if (possibilities.get(child).isFullName) {
+          sb.append(child);
+          names.add(sb.toString());
+        }
+      }
```
g-asweet: this is the kind of inefficiency that will come with the `isFullName` node design, you have to plausibly iterate overall the letters of the alphabet for each node which may or may not have fullNames associated. 

```java
for (String fullNames: possibilities.fullNames()) {
  // Add to names list
}
```

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -1,35 +1,122 @@
...
+    }
+    for (String letter : possibilities.keySet()) {
```
g-asweet: With the `isFullName` approach, aren't you not only iterating over letters at this point but also full name nodes?

capstone/src/test/java/com/google/sps/search/TrieTest.java
```diff
@@ -0,0 +1,194 @@
...
+    assertTrue(names.contains("Lucy Qu"));
+    assertTrue(names.contains("Lucy Liu"));
+    assertFalse(names.contains("Lucille Ball"));
+  }
+
+  /** Tests for search with prefix when one name in the trie is the substring of another name. */
+  @Test
+  public void searchWithPrefixOneNameIsSubstringTest() {
+    trie.insert("Qu", "Lucy Qu");
+    trie.insert("Quincy", "Jane Quincy");
```
g-asweet, line 208: I guess for thoroughness, you can also include Jane Squirrel

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/search/Trie.java
lxcyqx: Oh yeah I that bad naming then because the way I had it before the children could contain either single letter or full name. This solution still works for the way the code is written now, which is what you suggested -- having a set hold full names

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -1,35 +1,121 @@
...
+   * @param name user first name or last name
+   * @param fullName user full name.
+   */
   public void insert(String name, String fullName) {
     if (name.contentEquals("")) {
-      isEnd = true;
-      // add in the full name here
+      // indicates the end of first name or last name
+      this.isName = true;
+      fullNames.add(fullName);
```
g-asweet: instead of `.isName`, couldn't you just check if `!fullNames.isEmpty()`

g-asweet: I don't know if you want to expose `getFullNames()` or `getMatchingNames()` or whatever

g-asweet: The reason for this comment is to me it's usually a red flag when you have a bool that represents data being present, incase the boolean and the data are ever out of sync because of a bug. Any time you're representing data with 2 different things, I always try to reduce it down to the data itself.

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/search/Trie.java
lxcyqx: But if I do a !fullNames.isEmpty() check, then it would only add the first full name. If there are several full names that I want to add to the set, I wouldn't be able to add it since fullNames would not be empty after the first insertion

capstone/src/test/java/com/google/sps/search/TrieTest.java
```diff
@@ -0,0 +1,227 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Map;
+import java.util.Set;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class TrieTest {
```
nicorod-google, line 13: Thank you for these tests!

capstone/src/test/java/com/google/sps/search/SearchPredictorTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.mockito.Spy;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import com.google.appengine.tools.development.testing.LocalUserServiceTestConfig;
+import com.google.common.collect.ImmutableMap;
+
+public class SearchPredictorTest {
```
nicorod-google, line 23: I don't see any tests, so are you just setting this up to add tests later?

capstone/src/main/java/com/google/sps/search/Trie.java
nicorod-google: I think Andrew's idea is that when you check for this.isName on line 84 in findAll, you are just using the boolean to iterate through the fullNames set. The same can be achieved by performing the !fullNames.isEmpty() check. Therefore, you can use that to keep track of if this node contains a name instead of using the boolean. You could wrap this in a method i.e. isName() if you need to reference this outside of this file

capstone/src/main/webapp/scripts/search.js
```diff
@@ -0,0 +1,76 @@
...
+  name.innerHTML = suggested.fullname;
+  name.addEventListener("click", () => completeName(suggested.fullname));
+  suggestionsPanel.appendChild(name);
+}
+
+function completeName(fullname) {
+  searchInput.value = fullname;
+}
+
+autocomplete();
```
nicorod-google: Can you add this to the init method in group.js?

capstone/src/main/webapp/scripts/search.js
```diff
@@ -0,0 +1,76 @@
...
+function filter() {
+  const input = searchInput.value.toLowerCase();
+  suggestionsPanel.innerHTML = "";
+  const suggestions = names.filter(function(name) {
+    return name.fullname.toLowerCase().startsWith(input);
+  });
+  suggestions.forEach(suggested => appendSuggestion(suggested));
+  if (input === "") {
+    suggestionsPanel.innerHTML = "";
+  }
```
nicorod-google: Instead of clearing the suggestionsPanel if the input is empty at the end of the function, couldn't you just only add the suggestions if the input is not empty?

capstone/src/main/webapp/scripts/search.js
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function checkKey(e) {
+  let list = document.getElementById("suggestions");
+  if (list) {
+    list = list.getElementsByTagName("div");
+  }
```
nicorod-google: It is strange that you are reassigning the suggestions element to a list elements. Also, if "suggestions" does not exist, do you still want to pass that to the addActive function? I think you should split this into multiple variables and only continue with the key presses if "suggestions" exists

capstone/src/main/webapp/scripts/search.js
```diff
@@ -0,0 +1,76 @@
...
+  if (e.keyCode == "38") {
```
nicorod-google: I know you have comments but can you add functions for the different key codes that describe what that key press is? i.e. isUpArrow, isDownArrow, isEnter, etc.? It'd be easier to read that way and not rely on the comment since it is awkwardly below the conditional

capstone/src/main/webapp/scripts/search.js
```diff
@@ -0,0 +1,76 @@
...
+    currentFocus = list.length - 1;
+  }
+  list[currentFocus].classList.add("autocomplete-active");
+}
+
+function removeActive(list) {
+  for (let i = 0; i < list.length; i++) {
+    list[i].classList.remove("autocomplete-active");
+  }
+}
```
nicorod-google: This could be my misunderstanding, but why are you removing this class from all of the names in the list? Shouldn't only one name have the class? Could you find the element with the "autocomplete-active" class and then remove that class?

nicorod-google (changes requested):

capstone/src/main/webapp/scripts/search.js
g-asweet: ^ That or define constants at the top of the file:

```javascript
const UP_ARROW_KEYCODE = "38";
```

or

```javascript
const KeyCodes = {
  UP: "38",
  DOWN: ...
};
```

capstone/src/test/java/com/google/sps/search/SearchPredictorTest.java
lxcyqx: Yeah I'm going to add these later. I started writing tests but there were issues with mocking due to this object implementing serializable

capstone/src/main/webapp/scripts/search.js
lxcyqx: This is here because if I type a name it'll pull up results but then if I backspace all the way to nothing, the last posted suggestions are still going to be there, so I do this check to clear it. I tried doing what you suggested but the names still stay there

capstone/src/main/webapp/scripts/search.js
lxcyqx: edited in new pr to have different variables

capstone/src/main/webapp/scripts/search.js
lxcyqx: edited in new pr

capstone/src/main/webapp/scripts/search.js
lxcyqx: yeah you're right, I've edited in new pr

nicorod-google (approved):

PR #39, opened 2020-07-23
TF-IDF helper methods

himaniyadav:
Helper methods (and unit tests) for calculating TF-IDF
- `sanitize()`: Removes all punctuation and converts String to lowercase.
- `ngramTokenizer()`: Splits up a given string into n-grams, counting their frequency along the way

### DIFF
--- a/capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
+++ b/capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
@@ -14,6 +14,10 @@ public class TFIDFStringHelper {
    * Remove all punctuation and convert String to lowercase.
    */
   public static String sanitize(String input) {
+    if (input ==  null || input.length() == 0) {
+      return input;
+    }
+
     input = input.toLowerCase();
     // Regex identifies punctuation except apostrophes.
     input = input.replaceAll("[\\p{P}&&[^']]", "");
@@ -26,11 +30,18 @@ public class TFIDFStringHelper {
    * mapped with each term's freqency count.
    */
   public static LinkedHashMap<String, Integer> ngramTokenizer(String input) {
-    input = sanitize(input);
     LinkedHashMap<String, Integer> ngrams = new LinkedHashMap<String, Integer>();
+
+    if (input ==  null || input.length() == 0) {
+      return ngrams;
+    }
+
+    input = sanitize(input);
     String[] words = input.split(" ");
 
-    for (int n = 1; n <= 3; n++) {
+    int nLength = (words.length >= 3) ? 3 : words.length;
+
+    for (int n = 1; n <= nLength; n++) {
       for (int i = 0; i < words.length - n + 1; i++) {
         String ngram = concat(words, i, n);
         Integer count = ngrams.get(ngram);
--- a/capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
+++ b/capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
@@ -1,6 +1,7 @@
 package com.google.sps.Objects;
 
 import java.util.LinkedHashMap;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -44,12 +45,8 @@ public class TFIDFStringHelper {
     for (int n = 1; n <= nLength; n++) {
       for (int i = 0; i < words.length - n + 1; i++) {
         String ngram = concat(words, i, n);
-        Integer count = ngrams.get(ngram);
-        if (count == null) {
-          ngrams.put(ngram, 1);
-        } else {
-          ngrams.put(ngram, count + 1);
-        }
+        Integer count = ngrams.getOrDefault(ngram, 0);
+        ngrams.put(ngram, count + 1);
       }
     }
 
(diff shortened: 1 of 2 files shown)


capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
```diff
@@ -0,0 +1,72 @@
...
+   * Remove all punctuation and convert String to lowercase.
+   */
+  public static String sanitize(String input) {
+    if (input ==  null || input.length() == 0) {
+      return input;
+    }
+
+    input = input.toLowerCase();
+    // Regex identifies punctuation except apostrophes.
+    input = input.replaceAll("[\\p{P}&&[^']]", "");
```
g-asweet: Sanitization is always fun! I'd be careful of just replacing punctuation with no space, you could be creating new 1-grams this way. Instead I'd replace any you care about with a " " and then take a second pass that turns all 2-or-more-spaces in a row into 1 spaces, and remove all spaces at the beginning and end.

I'd also consider hyphens, dashes, periods, question marks, exclamation points --- do you want to find a valid NGram "here? I was" as "here i was", etc.

g-asweet: And then you have crazy cases like users posting urls "https://www.google.com/search?q=term&tbm=shop"

capstone/src/test/java/com/google/sps/objects/TFIDFStringHelperTest.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @Test
+  public void ngramTokenizer_notSanitized() {
+    assertEquals(TFIDFStringHelper.ngramTokenizer(testNgrams2), expectedNgramsMap2);
+  }
+
+  @Test
+  public void ngramTokenizer_twoWords() {
+    assertEquals(TFIDFStringHelper.ngramTokenizer(testNgrams3), expectedNgramsMap3);
```
g-asweet: Personally, I'd rather see the test data defined in the test, so I can very clearly see what you're testing. Especially since you're only using each of them once instead of referring to them in multiple tests.

capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
```diff
@@ -0,0 +1,72 @@
...
+   * Helper method for {@code ngramTokenizer()}. Concatenates words to form appropriate n-gram.
+   */
+  public static String concat(String[] words, int startIndex, int n) {
+    StringBuilder sb = new StringBuilder();
+    for (int i = startIndex; i < startIndex + n; i++) {
+      if (i > startIndex) {
+        sb.append(" ");
+      }
+      sb.append(words[i]);
+    }
```
g-asweet: One of these may work:

`String oneString = String.join(" ", myStringArray);`

or

`String oneString = String.join(" ", "first", "second", "third");`

g-asweet: Though I do like your usage of stringbuilder!

capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
```diff
@@ -0,0 +1,72 @@
...
+
+    input = sanitize(input);
+    String[] words = input.split(" ");
+
+    int nLength = (words.length >= 3) ? 3 : words.length;
+
+    for (int n = 1; n <= nLength; n++) {
+      for (int i = 0; i < words.length - n + 1; i++) {
+        String ngram = concat(words, i, n);
+        Integer count = ngrams.get(ngram);
```
g-asweet: `ngrams.getOrDefault(ngram, 1);`

g-asweet (dismissed):

capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Helper method for {@code sanitize()}. Removes urls from a String.
+   */
+  public static String removeUrls(String input) {
+    input = input.replaceAll("[\\S]+://[\\S]+", "");
+    input = input.replaceAll("www.[\\S]+", "");
+    return input;
+  }
```
nicorod-google, line 40: I think Andrew's comment wasn't specifically saying we need to sanitize URLs but rather giving an example of other punctuation that you would want to remove. From my understanding, this is the list of punctuation that will be removed: 
!"#$%&'()*+,-./:;<=>?@[]^_`{|}~
therefore I think what you have works and this isn't necessary. 

Also, I found the accepted answer and it's follow-up comments insightful: herehttps://stackoverflow.com/questions/18830813/how-can-i-remove-punctuation-from-input-text-in-java

nicorod-google: Though if you want to specifically remove URLs that is also fine :-)

nicorod-google (approved):

capstone/src/main/java/com/google/sps/Objects/TFIDFStringHelper.java
himaniyadav: Yes I saw that thread! I ended up going with a similar solution as found in the comments.

Also I do think removing URLs makes sense, so I think I'll leave it.

PR #41, opened 2020-07-24
TF-IDF Calculation

himaniyadav:
- Calculate the TF-IDF for all n-grams in all groups and generate 3 group tags.
- For now/for testing purposes, the servlet runs in the background every time the user's profile page loads.
- Not displayed on frontend yet, but these are examples of results stored in the database (tags highlighted so they're easier to see):
image
- The calculation appears to be correct as the words it identified are the most frequent - you can compare it with the [test data here](https://docs.google.com/spreadsheets/d/1HqcPHBz2LfFH3E49W9qbn8D9WLUFNQhjla0DawJioZw/edit?usp=sharing).
- Weighting according to Zipf's law (longer strings preferred) is not yet implemented, but that should help avoid repeats like "tribal," "tribal council," and "council."


nicorod-google (approved):

nicorod-google:
Make sure to add tests for the TagsIDFServlet in a follow-up PR

PR #42, opened 2020-07-25
Connect search to backend and results

lxcyqx:
- Ability to search for full name
- Connect search with backend
- Added full name field for querying. The full name field is all capital because datastore does not allow for case insensitive querying. 
- Add member to group upon clicking plus icon
- Display results from search upon pressing enter, clicking on name, or pressing enter after cycling through suggestions using arrow keys
- If user clicks on one of the suggested names, the results will list users with matching names first
- Update tries when new user is added to database
- Changed map in Trie to be from Character to Trie instead of String to Trie
- fix tests

Screen Shot 2020-07-24 at 11 56 26 PM

Screen Shot 2020-07-24 at 11 56 36 PM


capstone/src/main/java/com/google/sps/search/SearchPredictor.java
```diff
@@ -35,29 +39,81 @@ public SearchPredictor() {
...
+      } else {
+        int score = namesScore.get(name) + 1;
+        namesScore.put(name, score);
+      }
+
+      // If there is a complete name match for first or last name, increment score by 5
+      String[] split = name.split(" ");
+      partialName = partialName.toUpperCase();
+      if (partialName.equals(split[0]) || partialName.equals(split[1])) {
+        int score = namesScore.get(name) + FIVE;
```
himaniyadav: just curious, why five?

capstone/src/main/java/com/google/sps/search/servlets/NameTestDataServlet.java
```diff
@@ -18,14 +18,19 @@ public void init() {
...
-      String userId = firstName + "_" + lastName;
+      String fullName = firstName + " " + lastName;
+      String userId = firstName + "_" + lastName + counter;
       String email = userId + "@gmail.com";
-      User user = new User(userId, firstName, lastName, email, "", "", null, null, null);
+      User user =
+          new User(
+              userId, firstName, lastName, fullName.toUpperCase(), email, "", "", null, null, null);
       datastore.put(user.toEntity());
+      counter++;
```
himaniyadav: is the counter just to distinguish between duplicate names?

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
nicorod-google: +1 You should also rename this to something more descriptive i.e. COMPLETE_MATCH_SCORE

capstone/src/main/java/com/google/sps/search/servlets/NameTestDataServlet.java
lxcyqx: Yeah it is, I don't really need it for search but I just added it here

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
lxcyqx: I just needed a number that would increase the rank if either the first or last name matched completely. I had that if the first name prefix matched it would be 2 points and if last name prefix matched it would be 1 point. For example if the input is Andrew S, then Sasha Smith could be a possible option but I want names like Andrew Sweet, Andrew Brown etc to be ranked about Sasha Smith

himaniyadav: Okay, makes sense!

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
```diff
@@ -35,29 +39,81 @@ public SearchPredictor() {
...
+    for (String name : set) {
+      if (!namesScore.containsKey(name)) {
+        namesScore.put(name, increment);
+      } else {
+        int score = namesScore.get(name) + 1;
+        namesScore.put(name, score);
+      }
+
+      // If there is a complete name match for first or last name, increment score by 5
+      String[] split = name.split(" ");
```
nicorod-google: nit: Can you rename this to something like firstAndLastName, split doesn't tell me anything

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -79,16 +76,16 @@ public void insert(String name, String fullName) {
    * @return set of full names formed from node in Trie
    */
   private Set<String> findAll(Set<String> names, String prefix) {
-    Map<String, Trie> possibilities = children;
+    Map<Character, Trie> possibilities = children;
     /* If current node is the end of the a first name or last name, add the full name nodes to set of names to be returned. */
-    if (this.isName) {
+    if (!this.fullNames.isEmpty()) {
```
nicorod-google: You can now just use getIsName() here

capstone/src/main/java/com/google/sps/search/servlets/NameTestDataServlet.java
nicorod-google: Yup! This is just for test data but otherwise two people with the same would have the same ID

capstone/src/main/webapp/scripts/search.js
```diff
@@ -1,52 +1,104 @@
...
     addActive(list);
-  } else if (e.keyCode == "40") {
-    //down arrow
+  } else if (e.keyCode == KeyCodes.DOWN) {
     currentFocus++;
     addActive(list);
-  } else if (e.keyCode == 13) {
-    //press enter
+  } else if (e.keyCode == KeyCodes.ENTER) {
+    // e.preventDefault();
```
nicorod-google: nit: remove

capstone/src/main/webapp/scripts/search.js
```diff
@@ -1,52 +1,104 @@
...
+  } else {
+    getSuggestions();
+  }
+}
+
+function getSearchResults(list) {
+  let names = "";
+  if (list.length >= 5) {
+    //Gets top 5 suggested names
+    names = `${list[0].innerHTML},${list[1].innerHTML},${list[2].innerHTML},${list[3].innerHTML},${list[4].innerHTML}`;
```
nicorod-google, line 55: Curious why this is just the top 5 names?

capstone/src/test/java/com/google/sps/search/SearchPredictorTest.java
```diff
@@ -54,4 +56,10 @@ public void tearDown() {
     helper.tearDown();
     datastore = null;
   }
+
+  //  @Test
+  //  public void suggestTest() {
+  //    List<String> names = new ArrayList<String>();
+  //    when(searchPredictor.getNamesFromDb()).thenReturn(names);
+  //  }
```
nicorod-google: nit: remove

nicorod-google (changes requested):

capstone/src/main/webapp/scripts/search.js
lxcyqx: I don't want to have the results have too many names so I'm just limiting it to the top 5 names and finding results matching those top 5 names

nicorod-google: Okie dokie! A couple of things you might want to consider doing (as future changes, not in this PR):
1) If you're only using the top 5 names, then have the servlet only return the top 5 names.
2) Pagination for the names. Depending on how you do the sorting this might be less important feature but I can imagine a scenario in which a user searches for someone with a common name and their friend is not in the list of those 5 people. Pagination could be either serverside (you've already returned all the names so it's just using the pagination to display the next/previous set of 5) or clientside (how you did the comments in your portfolio)

nicorod-google (approved):

PR #47, opened 2020-07-29
Refine tfidf

himaniyadav:
- Refine the TF-IDF calculation by taking into account string length, number of words, and deduplicating words
- (To see results of "experiment" with changing around the above factors, see the [test data doc](https://docs.google.com/spreadsheets/d/1HqcPHBz2LfFH3E49W9qbn8D9WLUFNQhjla0DawJioZw/edit?usp=sharing))
- Add unit tests for the TagsTFIDFServlet
- Fix some edge cases, such as when there are no posts yet in a new group, etc


capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
```diff
@@ -144,21 +192,59 @@ public void calculateTFIDF(LinkedHashMap<Long, LinkedHashMap<String, Integer>> g
     }
   }
 
+  /**
+   * Weights the tf-idf score of a given ngram by taking into account its length.
+   */
+  private double weightScore(double score, String ngram) {
+    double wordsWeight = (ngram.split(" ").length / 3.1);
+    double lengthWeight = ((double) ngram.length() / longestNgramLength);
```
g-asweet, line 196: last thing we discussed was being able to weight these and they them out, so wordsWeight * wordsWeightScalar where scalar is a value [0, 1], and same for stringLengthScalar. You could run a simple eval where you set these values to a handful of different values, and figure out which fits best.

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
```diff
@@ -144,21 +192,59 @@ public void calculateTFIDF(LinkedHashMap<Long, LinkedHashMap<String, Integer>> g
...
-
+    
     ArrayList<Tag> topTags = new ArrayList<>();
     int tagsLength = (tagQueue.size() >= 3) ? 3 : tagQueue.size();
     for (int i = 0; i < tagsLength; i++) {
-      topTags.add(tagQueue.poll());
+      Tag next = tagQueue.poll();
+      if (next != null) {
+        if (checkIfDuplicate(topTags, next)) {
+          i--;
```
g-asweet, line 214: shouldn't this be a no-op instead? if you do i--, don't you end up checking the same value repeatedly by doing i--, and then in the for loop i++?

```java
if (!checkIfDuplicate(topTags, next)) {
  topTags.add(next);
}
```

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
```diff
@@ -144,21 +192,59 @@ public void calculateTFIDF(LinkedHashMap<Long, LinkedHashMap<String, Integer>> g
...
+      String[] existingWords = tag.getText().split(" ");
+      String[] nextWords = next.getText().split(" ");
+      for (String existingWord : existingWords) {
+        for (String word : nextWords) {
+          if (word.equals(existingWord)) {
+            return true;
+          }
+        }
+      }
+    }
```
g-asweet: instead of O(n**3), consider making a Set<String> of existing words, so you don't have to iterate over all existing terms and loop over them all. This could instead just be 

```java
for (String term : next.split(" ")) {
  if (existingTerms.contains(term)) {
    return true;
  }
}
return false;

**Review by g-asweet — CHANGES_REQUESTED** [review]

File: capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java (hunk 2 above)
Comment (g-asweet) [reply capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java:214]: I guess high level, I'm not sure if this is the direction you want

You're not allowing any overlap between any terms in ngrams, so `harry and hermione` and `potter and the` are invalid ngrams because they both use the word "and"

**Review by g-asweet — CHANGES_REQUESTED** [review]

File: capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java (hunk 2 above)
Comment (himaniyadav) [reply capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java:214]: No, earlier I call tagQueue.poll() which always retrieves and removes the first element in the queue.
The idea is that I want to loop through 3 times, and end up with three items added to the list. If I don't decrement i, I might end up with only one or two items in the list of topTags.

File: capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java [hunk 4]
```diff
@@ -62,39 +65,82 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-          (groupEntity.getProperty("posts") == null)
-              ? new ArrayList<Long>()
-              : (ArrayList<Long>) groupEntity.getProperty("posts");
-      List<String> postTexts = getGroupPostsText(postIds, response);
 
-      // TODO: Parallelize this process to occur for posts concurrently.
+      List<String> textData = getGroupPostsText(groupEntity, response);
+      addChallengeText(groupEntity, textData, response);
+
+      // TODO: Parallelize this process to occur for each string concurrently.
```
g-asweet, line 72: TECHNICALLY you probably want to batch them somehow, whether you just divide the array by 4 and assign the strings to up to 4 threads, or define some max strings per thread and dynamically determine threads based on number of strings.

Making a thread is very expensive and slow, so you need to find the right tradeoff between making thread and not making thread.

Currently, your test dataset isn't large enough to actually justify threading as a speed optimization, but it's still good practice to see if you could define the problem in the event that there's a million users

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
himaniyadav: Yeah, I tried that by changing the coefficients of the weight terms on this line:
`double totalWeight = 1 + wordsWeight + lengthWeight;`
It didn't really make much of a difference, besides changing up the orders of the top terms in a few cases. In basically every case it didn't cause the top terms to be any different, so I left it at 1. Although now that you point it out, I should probably change the weights to 0.5 for both instead.

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
himaniyadav: That's a good point about small words overlapping causing issues. I'm not really sure how else to best check for duplicate terms though. 
Currently for the case you described, if `harry and hermione` are added first because it has the highest score, and `potter and the` has the next highest score, `potter and the` would be counted as a duplicate and wouldn't be added, but `harry and hermione` would remain.

Maybe to mitigate against cases like that, I could say that if both terms I'm checking have two or more words, only mark it as a duplicate if two or more of those words are a match with the words in a term currently in the list. Here's what the results would be:

Currently in list  | Word to check | # Matches | Duplicate?
`harry and hermione` | `potter and the` | 1 | no
`and hermione` | `ron and harry` | 1 | no
`and hermione` | `and harry` | 1 | no
`harry and hermione` | `ron and hermione` | 2 | yes
`harry and hermione` | `harry and` | 2 | yes
`harry and` | `harry and hermione` | 2 | yes
`and hermione` | `ron and hermione` | 2 | yes
`harry and hermione` | `harry` | 1 | yes
`disney plus` | `disney` | 1 | yes
`disney` | `disney plus` | 1 | yes

The last 4 examples all involve an ngram with only one word so the logic wouldn't be changed from how it is now. Not sure if it's an ideal solution. I'm brainstorming other possibilities. Thoughts?

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
himaniyadav: This would work as an optimization if I didn't change any of my deduplication logic. If I did change the logic like how I showed in the comment above, then I wouldn't be able to use the set.

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
g-asweet: The answer MAY be not dedup-ing but adding another weight that measures duplicate results. You have your sliders of [0-1) that you can add a third one for "how duplicate" a string is?

I actually think I like your outcomes for the solution you listed though, I think all things I want to be no were no, and all things I want to be yes were yes.

You'd imagine in the `harry and hermione` | `ron and hermione` the winning ngram would be `and hermione`, due to it being significantly more prevelant than `harry and hermione` || `ron and hermione`

I think the relationship we're looking for is proportional, where `harry and hermione` should beat `and hermione` if it occurs > 2/3 of the time where 2 is the length of the 2gram and 3 is the length of the 3gram. Thoughts?

g-asweet: This would also mean that `Disney Plus` would only beat `Disney` if more than half of the time `Disney` was paired with `Plus`

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
g-asweet: Ack

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
g-asweet: (And if you end up with spare time between this and next week, I'd rather you implement a threading solution than leave it as a TODO :) )

capstone/src/test/java/com/google/sps/servlets/TagsTFIDFServletTest.java
```diff
@@ -0,0 +1,157 @@
...
+    ArrayList<Tag> tags = new ArrayList<>();
+    for (EmbeddedEntity tag : (ArrayList<EmbeddedEntity>) groupEntity.getProperty("tags")) {
+      tags.add(Tag.fromEntity(tag));
+    }
+    String tagsJson = new Gson().toJson(tags);
+
+    // expected score for "why hello" is tf * idf * weight =
+    // (1/3) * (natural log of 2) * (1 + 0.5(9/11) + 0.5(2/3.1)) = 0.40010108516...
+
+    assertTrue(tagsJson.contains("why hello") && tagsJson.contains("0.40010108516"));
```
g-asweet: This is going to be a pain to maintain if you were updating how you scored TF-IDF. I think this is fine, but consider using relative rankings as a way to test rather than absolute scores, where you can prove x > y > z is maintained. This will also more meaningfully break your test when you change how things are scored relatively, you'd have to update a test which returned `and hermione` instead of `harry and hermione`, allowing the test failure/update process itself to be useful to you, the developer.

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
himaniyadav: Yes, I was planning on adding this in my next PR!

capstone/src/main/java/com/google/sps/servlets/TagsTFIDFServlet.java
himaniyadav: Yeah, this step would happen after scoring and ranking so realistically `harry and hermione` and `ron and hermione` _probably_ wouldn't both show up. If they did, `and hermione` would definitely already be in the list, negating both of them. 

I think adding another "slider" wouldn't be very efficient since then I would have to do a similar "checkHowDuplicate()" calculation for every single ngram. Right now I'm only doing the dedup logic for the top handful of scores, but by introducing it as a weight I would have to calculate how duplicate a term is for hundreds or thousands of ngrams.

I think the proportional solution makes sense in theory but I am not sure how I could keep track of that. By the time I'm doing dedup I've already ranked everything by its frequency, length, etc. So just comparing the scores to get relative frequency wouldn't work because they're additionally weighted by n and the string length.

g-asweet (approved):

PR #49, opened 2020-07-29
Whitespace And Autocorrect

lxcyqx:
Whitespace:
Andrewsweet -> Andrew Sweet
Sweetandrew -> Andrew Sweet
Screen Shot 2020-07-29 at 5 11 59 PM
Screen Shot 2020-07-29 at 5 12 05 PM

Autocorrect: Maximum edit distance of two for each first name and last name. Algorithm combines recursion and dynamic programming

Screen Shot 2020-07-29 at 5 13 57 PM

Ordering: 
Ordering based on match. Names with lower LED distances from target word are listed first

Front end changes:
Limit to 5 suggestions but now results page will list all possible users
Screen Shot 2020-07-29 at 5 24 54 PM


g-asweet:
That's a lotta Yubaba
> Screen Shot 2020-07-29 at 5 24 54 PM

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+      }
+      if (this.hasFullName(firstName, fullName)) {
+        fullNames.add(fullName);
+      }
+    }
+    return fullNames;
+  }
+
+  /**
+   * Finds names that are within the maximum LED distance of input by calling a recursive helper
```
g-asweet: Can you clarify LED in the comment?

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+      Map<String, Integer> suggestions) {
+    int numColumns = input.length() + 1;
+    List<Integer> currRow = new ArrayList<Integer>();
+    currRow.add(prevRow.get(0) + 1);
+
+    for (int j = 1; j < numColumns; j++) {
+      if (input.charAt(j - 1) == currChar) {
+        currRow.add(prevRow.get(j - 1));
+      } else {
+        currRow.add(1 + Math.min(Math.min(prevRow.get(j), currRow.get(j - 1)), prevRow.get(j - 1)));
```
g-asweet: `Math.min(Math.min(prevRow.get(j), currRow.get(j - 1)` can you do this one line above, and give it a good variable name? Hard for me to follow as is.

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+   * @param input user input
+   * @param prevRow previous row in matrix
+   * @param suggestions map of suggested names
+   */
+  private void searchLedRecursive(
+      Trie currNode,
+      Character currChar,
+      String input,
+      List<Integer> prevRow,
+      Map<String, Integer> suggestions) {
```
g-asweet: As defined, this is hard for me to follow, I can't clearly see where the relationship exists between currNode, currChar, input, prevRow, Suggestions, and edit distance overall... I'm wondering if some kind of a data structure could help, or more contextual variable names? I'll spend some time reading, but at a glance, it's not immediately clear what's happening to me

g-asweet: Consider returning the suggestions map rather than taking one in? This will at least clarify the input/output relationship a bit.

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+  /**
+   * Finds names that are within the maximum LED distance of input by calling a recursive helper
+   * method that uses dynamic programming to find the edit distance between words.
+   *
+   * @param input user input
+   * @return Map from possible names to its edit distance from input
+   */
+  public Map<String, Integer> searchLed(String input) {
+    input = input.toUpperCase();
+    Map<String, Integer> suggestionsMap = new HashMap<String, Integer>();
```
g-asweet, line 142: Either move this closer to the first usage of this variable, or remove it entirely, instead returning and merging suggestionsMaps, recursively.

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+      } else {
+        currRow.add(1 + Math.min(Math.min(prevRow.get(j), currRow.get(j - 1)), prevRow.get(j - 1)));
+      }
+    }
+
+    /*
+     * If current row indicates end of name and is less than or equal to max LED, add all full names
+     * to map of suggestions and indicate edit distance from input.
+     */
+    int ledDistance = currRow.get(currRow.size() - 1);
```
g-asweet: `editDistance` or `ledEditDistance` would help make this make a bit more sense.

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+    if (ledDistance <= MAX_LED && currNode.getIsName()) {
+      for (String name : currNode.fullNames) {
+        suggestions.put(name, ledDistance);
```
g-asweet: here you have a map called suggestions which is mapping names to editDistances...

I this is maybe where some of the naming comes in. ledDistance is bigger the more edits away I assume, and so your scoring would be the opposite of my expectation (a higher score would indicate a worse suggestion, right?)

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+
+  /**
+   * Recursive helper method that populates the new row with values indicating edit distance with
+   * input name. These values are based off of the subproblem (the review row that is passed in).
+   *
+   * @param currNode current trie
+   * @param currChar current character
+   * @param input user input
+   * @param prevRow previous row in matrix
+   * @param suggestions map of suggested names
```
g-asweet: Map of suggested names to what?

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+  }
+
+  /**
+   * Recursive helper method that populates the new row with values indicating edit distance with
+   * input name. These values are based off of the subproblem (the review row that is passed in).
+   *
+   * @param currNode current trie
+   * @param currChar current character
+   * @param input user input
+   * @param prevRow previous row in matrix
```
g-asweet: Previous row of what are these integers? prevEditDistances?

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+    return suggestedNames;
+  }
+
+  /**
+   * Recursive helper method that populates the new row with values indicating edit distance with
+   * input name. These values are based off of the subproblem (the review row that is passed in).
+   *
+   * @param currNode current trie
+   * @param currChar current character
+   * @param input user input
```
g-asweet: `input user input` is not particularly helpful for context, consider `searchString` `nameString` `queryString`

something like `The user input string representing the search for a user by first and/or last name`

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+
+    return suggestedNames;
+  }
+
+  /**
+   * Recursive helper method that populates the new row with values indicating edit distance with
+   * input name. These values are based off of the subproblem (the review row that is passed in).
+   *
+   * @param currNode current trie
+   * @param currChar current character
```
g-asweet, line 162: These are probably fine as is, but consider adding what the current trie/char represents? All previous characters are correct? We're already possibly some existing edit distance away?

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+      for (String name : currNode.fullNames) {
+        suggestions.put(name, ledDistance);
+      }
+    }
+
+    /*
+     * If any entry in the current row are less than or equal to the maximum allowed LED
+     * distance, recursively search each branch of the trie from current character.
+     */
+    int minRowVal = Collections.min(currRow);
```
g-asweet: `minRowVal` is also not as helpful, `minEditDistance` or `minEditDistanceForRow` or equivalent?

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
+    if (minRowVal <= MAX_LED) {
+      for (Character c : currNode.children.keySet()) {
+        this.searchLedRecursive(currNode.children.get(c), c, input, currRow, suggestions);
+      }
+    }
+  }
+
+  /**
+   * Helper method that checks whether the trie contains the full name by recurring down the tree
```
g-asweet: `by recursing`

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
```diff
@@ -116,7 +147,52 @@ public int compare(Map.Entry<String, Integer> a, Map.Entry<String, Integer> b) {
...
+      lower score */
+      if (nameToLedMap.get(name) != 0) {
+        score = 1 / nameToLedMap.get(name);
+      }
+      if (namesScore.containsKey(name)) {
+        namesScore.put(name, namesScore.get(name) + score);
+      } else {
         namesScore.put(name, score);
       }
     }
```
g-asweet: This is clever and I like it (addToMap + the variations) However, the absense of the `double increment` parameter now triggers separate reciprocal logic, so I would be careful overloading the same method name `addToMap` since the behavior is actually pretty different. I would instead call this something like `addToMapWithEditDistance` or something that clearly indicates that this logic is unique.

capstone/src/main/java/com/google/sps/search/Trie.java
```diff
@@ -93,6 +96,140 @@ public void insert(String name, String fullName) {
...
   }
 
+  /**
+   * Checks whether the input can be split into a full name by splitting input at each index and
+   * checking whether the newly created first and last name exist in the set of full names in the
+   * trie. For example, input "andrewsweet" can be split in "Andrew Sweet," and if this name is in
+   * the trie, then the name will be added to the set. Boolean reversed indicates whether to search
+   * for the name but first and last name are reversed. For example, "sweetandrew" is reversed and
+   * it can be split but into "Sweet Andrew," for which we can check that the reverse ("Andrew
+   * Sweet") is in the trie.
```
g-asweet, line 106: This is very robust! Nice work!!

capstone/src/main/java/com/google/sps/search/SearchPredictor.java
```diff
@@ -17,7 +17,10 @@
 
 public class SearchPredictor implements Serializable {
 
-  private static final int COMPLETE_PARTIAL_NAME_MATCH = 5;
+  private static final double COMPLETE_PARTIAL_NAME_MATCH = 5;
+  private static final double COMPLETE_MATCH = 10;
+  private static final double FIRST_NAME_MATCH = 2;
+  private static final double LAST_NAME_MATCH = 1;
```
g-asweet, line 17: Can you add tests for comparative scoring? COMPLETE_MATCH result ranked above PARTIAL_NAME above FIRST above LAST

g-asweet: An additional test could compare 0 edit distance, 1 edit distance, 2 edit distance, and if you wanted confirm that >= 3 edit distance does not return a result.

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/search/Trie.java
lxcyqx: Yep, so this is why I did the reciprocal of the edit distance when adding it to the main name to score map in SearchPredictor, which you commented on in one of the below comments

g-asweet (approved):
This is much easier to follow now :D

nicorod-google (approved):

PR #53, opened 2020-07-31
Group tags front end: show context/usage

himaniyadav:
- When clicking a tag, a modal opens that shows the user where those words were used in the context of the Group's posts, comments, and challenges.

image
image

**Additional note**: 
I used the Google Cloud Scheduler API to schedule a cron job that calls the tags-tfidf servlet every 24 hours. However, that is only called on the deployed version, so I've left the function commented out in `profile.js` for now, in case we need to uncomment and run the calculation the dev server for testing purposes.

### DIFF
--- a/capstone/src/main/webapp/scripts/tags.js
+++ b/capstone/src/main/webapp/scripts/tags.js
@@ -16,9 +16,13 @@ function displayTags(tags) {
       tagContainer.appendChild(text);
   } else {
     for (tag of tags) {
+      let tagName = tag.text;
       const tagElement = document.createElement('span');
       tagElement.id = 'tag';
-      tagElement.innerText = tag.text;
+      tagElement.innerText = tagName;
+      tagElement.addEventListener("click", function () {
+        openTagsContext(tagName);
+      });
       tagContainer.appendChild(tagElement);
     }
   }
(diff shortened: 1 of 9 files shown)


capstone/src/main/webapp/scripts/tags.js
```diff
@@ -33,3 +37,41 @@ function openAboutTags() {
...
+    textElement.id = 'context-text';
+    const boldedText = boldTag(context.text, context.tagText);
+    textElement.innerHTML = `${context.type}:	${boldedText}`;
+    container.appendChild(textElement);
+  }
+}
+
+/** Bold the given part of a text. */
+function boldTag(text, tag) {
+  return text.replace(new RegExp('(^|\\W)(' + tag + ')(\\W|$)','ig'), '$1<strong>$2</strong>$3');
```
g-asweet, line 76: I like this approach! Double checking, would this allow for multiple occurrences to be bolded in the same string? From what I can tell, you've marked this as case insensitive AND global with the 'ig', so I think the answer should be yes.

g-asweet (approved):

capstone/src/main/webapp/scripts/tags.js
himaniyadav: Yes, it will!

PR #59, opened 2020-08-04
Graph

lxcyqx:
Complete user ranking by dynamically building graph while using Dijkstra's algorithm to find the shortest path to target user

### DIFF
--- a/capstone/src/main/java/com/google/sps/graph/Dijkstra.java
+++ b/capstone/src/main/java/com/google/sps/graph/Dijkstra.java
@@ -8,33 +8,38 @@ import java.util.Set;
 public class Dijkstra<V extends Vertex<E>, E extends Edge<V>> {
 
   public Vertex<E> runDijkstra(V src, V dest) {
-    // used to keep track of vertices closest to the source vertex
+    System.out.println("running dijkstra");
+    // Used to keep track of vertices closest to the source vertex
     PriorityQueue<V> minHeap = new PriorityQueue<V>(new VertexComparator<E>());
     src.setDistance(0);
     minHeap.add(src);
     Set<V> visitedSet = new HashSet<V>();
     while (!minHeap.isEmpty()) {
-      // pop off vertex with smallest distance from src
+      // Pop off vertex with smallest distance from src
       V current = minHeap.peek();
       visitedSet.add(minHeap.poll());
+      System.out.println("current " + current.getId() + "dest " + dest.getId());
+      System.out.println("visited set " + visitedSet);
       if (current.equals(dest)) {
+        System.out.println("current equals dest----------------------");
         return current;
       }
       List<E> outgoingEdges = current.getOutgoingEdges();
+      System.out.println("outgoing edges dijkstra " + outgoingEdges);
       if (outgoingEdges == null) {
         return null;
       }
       for (E e : outgoingEdges) {
-        // get the vertex on the other side of the edge
+        // Get the vertex on the other side of the edge
         V adjacent = e.getDestVertex();
-        // update the shortest path to adjacent vertex
+        // Update the shortest path to adjacent vertex
         double newDistanceToAdjacent = current.getDistance() + e.getEdgeWeight();
         if (newDistanceToAdjacent < adjacent.getDistance()
             || (!minHeap.contains(adjacent) && !visitedSet.contains(adjacent))) {
           adjacent.setDistance(newDistanceToAdjacent);
           // remember path to this vertex
           adjacent.setPrev(e);
-          // trigger reheapify since vertex has been mutated
+          // Trigger reheapify since vertex has been mutated
           if (minHeap.contains(adjacent)) {
             minHeap.add(minHeap.remove());
           } else {
@@ -43,6 +48,7 @@ public class Dijkstra<V extends Vertex<E>, E extends Edge<V>> {
         }
       }
     }
+    System.out.println("returning dest ---------------------" + dest.getDistance());
     return dest;
   }
 }
--- a/capstone/src/main/java/com/google/sps/graph/Dijkstra.java
+++ b/capstone/src/main/java/com/google/sps/graph/Dijkstra.java
@@ -1,0 +1,48 @@
+package com.google.sps.graph;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.PriorityQueue;
+import java.util.Set;
+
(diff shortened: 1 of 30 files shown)


capstone/src/main/java/database/DatabaseRetriever.java
```diff
@@ -0,0 +1,102 @@
+package database;
```
nicorod-google: Why did you move this from java/com/google/sps/search/ to java/database? We should keep everything under the java/com/google structure

capstone/src/main/java/com/google/sps/search/servlets/SearchResultsServlet.java
```diff
@@ -26,21 +35,32 @@
 @WebServlet("/search-results")
 public class SearchResultsServlet extends AuthenticatedServlet {
 
+  private static double COMPLETE_MATCH = 10;
```
nicorod-google: nit: make this final

capstone/src/main/java/database/DatabaseRetriever.java
```diff
@@ -0,0 +1,102 @@
...
+import com.google.appengine.api.datastore.EntityNotFoundException;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.sps.graph.UserVertex;
+
+public class DatabaseRetriever implements Serializable {
+
+  /** */
+  private static final long serialVersionUID = 1L;
```
nicorod-google: I'm not sure this is used anywhere? If it's necessary, then please fill out the comment because I don't know what it does as well as make it a constant since it's final.

nicorod-google (changes requested):

capstone/src/main/java/com/google/sps/graph/VertexComparator.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class VertexComparator<E> implements Comparator<Vertex<E>> {
+
+  @Override
+  public int compare(Vertex<E> v1, Vertex<E> v2) {
+    if (v1.getDistance() < v2.getDistance()) {
+      return -1;
+    } else if (v1.getDistance() > v2.getDistance()) {
+      return 1;
+    }
```
g-asweet:
```java
return Double.compare(v1.getDistance(), v2.getDistance());
```

capstone/src/main/java/com/google/sps/graph/Dijkstra.java
```diff
@@ -0,0 +1,48 @@
...
+          // remember path to this vertex
+          adjacent.setPrev(e);
+          // Trigger reheapify since vertex has been mutated
+          if (minHeap.contains(adjacent)) {
+            minHeap.add(minHeap.remove());
+          } else {
+            minHeap.add(adjacent);
+          }
+        }
+      }
```
g-asweet: I think you should move 

```java
for (E e : outgoingEdges) {
  // Get the vertex on the other side of the edge
  V adjacent = e.getDestVertex();
  if (!visitedSet.contains(adjacent)) {
     ...
  }
}
```

to wrap all of this, as you never want to visit the same point twice.

g-asweet (changes requested):

capstone/src/main/java/com/google/sps/graph/Dijkstra.java
g-asweet: I'd pull this out into a `processNeighbors(current, visitedSet, minHeap)` method

g-asweet: Unless you want to share a link to whichever resource you found online which most helped in shaping your dijkstra implementation, which I coudl be wrong about

g-asweet: https://www.vogella.com/tutorials/JavaAlgorithmsDijkstra/article.html Hmm this seems to be pretty close to what you have

capstone/src/main/java/com/google/sps/search/servlets/SearchResultsServlet.java
```diff
@@ -50,13 +70,55 @@ public void doGet(String userId, HttpServletRequest request, HttpServletResponse
       for (Entity result : pq.asIterable()) {
         try {
           users.add(User.fromEntity(result));
+          namesScore = this.runDijkstra(namesScore, userId, searchString, User.fromEntity(result));
```
g-asweet, line 72: HIGH level, I'm concerned that every time somebody types in a new character, you're getting a list of names using your trie implementation AND following up that entire result list with a call to dijkstra for every user in the result list......

If you're going to go a pathfinding route, I would suggest precalculating and saving distances to the database, or at least caching results.

capstone/src/main/java/database/DatabaseRetriever.java
g-asweet: I believe this has something to do with being a serializable object.

https://www.baeldung.com/java-serial-version-uid#serial-version-uid

If you're saving data into some place, and then updating your logic, and trying to load that serialized data back, you could end up in a mess if they objects weren't the same. This versioning allows you to identify those changes.

g-asweet (dismissed):
Last comment is that this is a lot of change for the second to last day of your internship. I'd take a crack at making it work today, but I wouldn't stress over making sure you've submitted this. Documenting your existing features and future plans of solving a graph problem will do you just as well as finishing this.

capstone/src/main/java/com/google/sps/search/servlets/SearchResultsServlet.java
lxcyqx: This servlet is only called when the user presses enter to list the user results, so it'll only do the pathfinding once when the user presses enter and not at each new character. Caching would still be useful though

capstone/src/main/java/database/DatabaseRetriever.java
lxcyqx: I just created a new package and it defaulted to that. I've now changed it to be under the java/com/google structure. Is there a reason for it to be under that though?

nicorod-google: That's the structure for java code at Google: 
java/com/google/a/b/c/ClassName.java if it will be part of any production system

nicorod-google (approved):

PR #61, opened 2020-08-05
Profile Pictures 

anikabagga:
- Added ability to upload a profile picture to user profile 
- Group posts, comments, members updated to show profile picture (if uploaded)

Screenshot 2020-08-04 at 11 01 45 PM
Screenshot 2020-08-04 at 10 56 48 PM


capstone/src/main/java/com/google/sps/servlets/ParseTestGroupDataServlet.java
```diff
@@ -80,14 +80,16 @@ private void createGroupEntity(String[] fields) {
    * Create a Post object and add it to the datastore given information from TSV file.
    */
   private long createPost(String postText, String author) {
+    String authorId = author;
     String authorName = author;
```
nicorod-google, line 84: Why is the authorId the same as authorName?

capstone/src/main/webapp/scripts/profile.js
```diff
@@ -196,36 +202,21 @@ function closeModal(type) {
...
-    fetch("/editProfile", { method: "POST", body: params });
+  // Add user profile picture if previously uploaded 
+  if (user.profilePic != null & user.profilePic != "") {
+    let profilePicDiv = document.getElementById("modal-profile-pic");
+    let imageContent = document.createElement("img");
+    imageContent.className = "profile-pic-img";
+    imageContent.src = "serve?blob-key=" + user.profilePic;
+    profilePicDiv.appendChild(imageContent);
   }
 }
```
nicorod-google, line 221: Are you sure we want to delete this and it's not used elsewhere?

nicorod-google (changes requested):
Why is the EditProfileServletTest being deleted? The servlet still exists, so I would think we still want those tests?

capstone/src/main/java/com/google/sps/servlets/ParseTestGroupDataServlet.java
himaniyadav: Initially when I wrote the ParseTestGroupDataServlet I set authorName to be equal to the userId - I thought for the test data it was okay if the post displayed "JANE_DOE" instead of "JANE DOE". In the TSV file for group data we only have userIds to identify the members of a group, but no other user information. If we wanted to make the test data more accurate we could split up the name by "_" and set the author name that way. In the real post servlet we'll have access to all the user information so they would be different.

capstone/src/main/webapp/scripts/profile.js
himaniyadav: ^Yes I think deleting saveEdits() might have been a mistake? Or if not, I'm not sure why we're removing the ability to save edits to a profile?

himaniyadav: Oops nevermind, I see you did it in the html instead. In that case it's probably fine because it wasn't used elsewhere, as long as saving edits to the profile still works and all

capstone/src/main/webapp/profile.html
```diff
@@ -52,7 +52,9 @@ <h1 id="ongoing-challenges">Ongoing Challenges</h1>
         <span class="close" onclick="closeModal('edit')">&times;</span>
         <div id="modal-profile-pic"></div>
         <h2 id="modal-name-container"></h2>
-        <form id="edit-profile">
+        <form id="edit-profile" method="POST" enctype="multipart/form-data">
```
himaniyadav, line 55: I tested this out and it works, but I'm sort of confused why? I'm not super familiar with html forms but I thought there had to be an action="\editProfile" or whatever servlet you want it sent to.

capstone/src/main/webapp/profile.html
```diff
@@ -76,7 +78,7 @@ <h2 id="modal-name-container"></h2>
             <input type="text" class="input-field" placeholder="Add an interest" 
               name="interests" id="interests">
           </li>          
-          <button class="save-btn" onclick="saveEdits()">SAVE</button>
+          <button class="save-btn">SAVE</button>
```
himaniyadav, line 81: Same as above. I tested this out and it works, but I'm sort of confused why? I'm not super familiar with html forms but I thought the save button would need to have a type="submit"

capstone/src/main/webapp/profile.html
anikabagga: Forms are a bit different with Blobstore. The action of the form becomes a URL that points to Blobstore. When the form is submitted, the request first goes to Blobstore that takes care of handling the file uploaded by the user and then forwards the request to the servlet (which can get all the other data inputed in the form).

capstone/src/main/webapp/profile.html
anikabagga: Blobstore handles forms differently. Explained above.

capstone/src/main/java/com/google/sps/servlets/ParseTestGroupDataServlet.java
anikabagga: I set the authorId and authorName to the same value since our test group data only has the authorId's. This is a just a workaround for the test data, but the correct authorId and authorName will be stored for a real user.

capstone/src/main/webapp/profile.html
himaniyadav: Oh interesting, thanks!

capstone/src/main/java/com/google/sps/servlets/ParseTestGroupDataServlet.java
nicorod-google: Ah I didn't notice this was for the test data, thank y'all for the explanation!

capstone/src/main/webapp/scripts/profile.js
nicorod-google: Interesting, totally missed that. I also didn't know that's how forms needed to work for blobstore, thanks for the explanation!

nicorod-google (approved):

